Issue #117, opened 2012-03-11
paper_trail breaks globalize3, even without :versioning => true


asmolero:
This is the set up I have in my model:

``` ruby
translates :name, :description
has_paper_trail
```

Changes in `name` and `description` are never reflected to the database when `has_paper_trail`

I have checked the 4 cases and this is the result: https://gist.github.com/2015214

_has_paper_trail ON and :versioning => true_: Doesn't work
_has_paper_trail ON and :versioning => false_: Doesn't work
_has_paper_trail OFF and :versioning => true_: Works but paper_trail is disabled for the model attributes
_has_paper_trail OFF and :versioning => false_: It works, but no versioning in the translations nor in the model

PR #241, opened 2013-06-16
Unit test for a case described in issue #117 (the changes ware not saved if the model used both paper_trail and globalize3)

kozak:
This has been already *fixed* in the newest (github) version of globalize3, but maybe that test will be useful. 

For reference: 
This was failing for me on gem 'globalize3'; Works on gem from github. Also this workaround fixes it: 

```ruby
module PaperTrail
  module Model
    module InstanceMethods

      def initialize_copy(source)
        obj = super(source)
        obj.tap { |o| o.send(:remove_instance_variable, :@globalize) } rescue obj
        obj
      end

    end
  end
end
```

### DIFF
--- a/test/data/models/dish.rb
+++ b/test/data/models/dish.rb
@@ -1,0 +1,9 @@
+require 'paper_trail'
+
+class Dish < ActiveRecord::Base
+  belongs_to :restaurant
+  
+  translates :name
+  
+  has_paper_trail
+end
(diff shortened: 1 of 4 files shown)


scarfacedeb:
@kozak Is it still an issue with (globalize-versioning)[https://github.com/globalize/globalize-versioning]?
@parndt Maybe we should transfer those tests there instead.

parndt:
Yes, this should go in globalize-versioning

PR #294, opened 2013-11-02
Removed Hash[array] in favour of Enumerable#reduce({})

parndt:
Just fiddling with improving the code quality and readability.

If this is not the case with my refactorings (i.e. they're only easier for me), let me know.


scarfacedeb:
@parndt @shioyama do you still consider merging it?
I like the changes, especially inside `act_macro.rb`.
It's a bit easier to understand what `globalize/translates` method does, because `each` loop was refactoring into semantically named method.

simi:
This is really old. I think some improvements can be applied. But act_macro is patched already with similar patches.

scarfacedeb:
You're right, I forgot to diff it with the latest changes.

shioyama:
@scarfacedeb @simi I'm okay with this if someone is willing to update it. Although it means an API change (`translates` -> `globalize`) so we should add a deprecation message at least.

parndt:
Wow, I can't believe I wrote all that code.. I still really like the changes.

parndt:
@shioyama the deprecation and alias already happens: b64ff9c99e509c0e0808ca8ba1a934576a30ace7

simi:
@parndt https://github.com/globalize/globalize/pull/294#issuecomment-74694712

PR #328, opened 2014-02-04
Don't create translation record for default_locale on save

scarfacedeb:
I want to allow users to create localized pages for a specific locale, excluding default locale.
That way I can have a czech-specific page only on the czech version of a site.
But I can't do that because when I'm trying to save a new `Page` record with a `Page::Translation` association (`locale == :cz`), rails gives me a validation error, because globalize has added another association with `locale == :en`.

I'm using `rails_admin` and its default field for has_many associations. 

I think that my problem starts  inside `save` method.

``` ruby
      def save(*)
        Globalize.with_locale(read_attribute(:locale) || I18n.default_locale) do
          super
        end
      end
```

In particular, this part in `read_attribute`:

``` ruby
        if name == :locale
          self.try(:locale).presence || self.translation.locale
```

Which leads to `translation` method:

``` ruby
translation_for(::Globalize.locale)
```

As a result, `translation_for` builds an empty translation for default `::Globalize.locale` locale.

---

I think that `read_attribute(:locale)` could skip creating default translation record, because in that particular case it only uses it to get its locale, that is equal to `Globalize.locale` anyway.
I ran `rake test` on this branch and on the branch with merged _before_callbacks_ (it's called _combined_) and there's no new errors.

### DIFF
--- a/lib/globalize/active_record/instance_methods.rb
+++ b/lib/globalize/active_record/instance_methods.rb
@@ -46,7 +46,7 @@ module Globalize
         return super(name) unless options[:translated]
 
         if name == :locale
-          self.try(:locale).presence || self.translation.locale
+          self.try(:locale).presence || Globalize.locale
         elsif self.class.translated?(name)
           if (value = globalize.fetch(options[:locale] || Globalize.locale, name))
             value
--- a/test/data/models/artwork.rb
+++ b/test/data/models/artwork.rb
@@ -1,0 +1,4 @@
+class Artwork < ActiveRecord::Base
+  translates :title
+  accepts_nested_attributes_for :translations
+end
--- a/test/data/schema.rb
+++ b/test/data/schema.rb
@@ -221,4 +221,13 @@ ActiveRecord::Schema.define do
     t.references :post
     t.string :file_type
   end
+
+  create_table :artworks, :force => true do |t|
+  end
+
+  create_table :artwork_translations, :force => true do |t|
+    t.string     :locale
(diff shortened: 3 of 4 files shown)


shlima:
+1

jerefrer:
Any news on this ?

I'm having the same issue and I have trouble imagining how there could not be an army of people effected by this behavior.
Also it took me about an hour and a half of painful debugging and questioning to arrive to the conclusion that something might be wrong with globalize itself.

Looking at the Travis failure it seems to me that it's unrelated.

Would you be so kind as to merge this ?
Thanks !

jtomaszewski:
@jerefrer , I'm going through all of it now as well.

> I'm realizing that the reason I have to fight globalize might be that empty translations are supposed to happen, as pardnt was saying in another issue. I can't take more time on this for today, I'll update with future realizations tomorrow.

You mean this comment https://github.com/globalize/globalize/pull/195#issuecomment-13008495 ?

Well, we should start with that question then: Do we want to create a Translation object for a default_locale even if we don't use it at all (f.e. it doesnt have any attributes set)?

My cases:

- I want to create a record which doesn't have any translations
- I want to create a record which has translations, but none of them are in default_locale

jtomaszewski:
@scarfacedeb 's fix is working and not really messing up much with the specs. I rebased it in https://github.com/globalize/globalize/pull/578 so you can see it. How about somebody gives us a decision what we're gonna do finally about it? ;)

PR #330, opened 2014-02-11
Set touch: true on belongs_to globalized_model

shlensky:
I think that it is right (sorry for my bad english)

### DIFF
--- a/lib/globalize/active_record/class_methods.rb
+++ b/lib/globalize/active_record/class_methods.rb
@@ -44,7 +44,7 @@ module Globalize
             klass = self.const_set(:Translation, Class.new(Globalize::ActiveRecord::Translation))
           end
 
-          klass.belongs_to :globalized_model, :class_name => self.name, :foreign_key => translation_options[:foreign_key]
+          klass.belongs_to :globalized_model, :class_name => self.name, :foreign_key => translation_options[:foreign_key], :touch => true
           klass
         end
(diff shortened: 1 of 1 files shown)


heathd:
@shioyama have you had any further thoughts relating to this? We are using globalize and also [being affected by affected by this issue](https://github.com/alphagov/whitehall/commit/95ee80a05018d805e91ca3505596739a481bd040). For the time being we will likely need to fork globalize to work around this issue which we'd prefer not to do.

It feels to me that a better fix for @shlensky original issue might be to modify override `#cache_key` for globalised models to take into account the updated_at timestamp of any translated part.

shioyama:
@heathd I agree we'd prefer not to have people forking globalize because of this issue. We can revert that change and release a new version, and look into `cache_key` as an alternative.

@parndt @simi What do you think?

simi:
I think we can accept a patch to enable/disable this.

```ruby
translates :meta, touch: true #=> (default to false)
```

heathd:
I've opened a PR #408 for discussion/review which changes the implementation of `#cache_key`

krisdigital:
The problem with this solution is that it does not work if your translated model belongs to another model connected with `touch: true`.. Something that worked before.. Now only the translated model itself is cached properly, which makes the `touch: true` option kind of useless. Or did I get it wrong?

PR #340, opened 2014-03-05
Fix save issue when updating multiple locales at once

graudeejs:
We had update record issues in our admin interface.

We basically update title_en, title_ru, title_lv with single form (using globalize-accessors).
In various conditions this used to fail, because of line 54.

I wanted to update globalize and globalize-accessors to be able to track changes for each locale, but didn't find easy way to do this in reasonable amount of time.

Anyway, what do you think about this patch?
Is there anything I can/need to do to get it merged?

If you need more detailed example of failing scenario let me know.

### DIFF
--- a/lib/globalize/active_record/instance_methods.rb
+++ b/lib/globalize/active_record/instance_methods.rb
@@ -44,21 +44,17 @@ module Globalize
             options = {:locale => options}
           end
           options = {:locale => Globalize.locale}.merge(options)
-          
+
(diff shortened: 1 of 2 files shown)


scarfacedeb:
@parndt I suppose we can close it, because #341 was merged and it makes it possible to use `accepts_nested_attributes_for` with translations without any problem.

@graudeejs is it still an issue?

parndt:
Reopen if so

Issue #325, opened 2014-01-26
Is there a way to use accepts_nested_attributes_for with null: false columns?


scarfacedeb:
I'm working on a plugin for rails_admin: https://github.com/scarfacedeb/rails_admin_globalize_field and I'm having an issue with `null: false` columns (without default value) in translations tables.

I can successfully add/update records from rails_admin, because I'm submitting translations using nested attributes (e.g. artwork[translations][title]) and activerecord process them as any ordinary association.

But now I can't save the records using generated globalize methods. (e.g. artwork.title=; see example below)

Relevant migration code:

``` ruby
Artwork.create_translation_table! title: { type: :string, null: false }
```

Model code:

``` ruby
translates :title
accepts_nested_attributes_for :translations, allow_destroy: true
```

And console output:

``` ruby
> Artwork.create({ title: "test", category_id: 1, image: "file.jpg"  })
(0.2ms)  BEGIN
SQL (0.6ms)  INSERT INTO "artworks" ("category_id", "created_at", "image", "slug", "updated_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"  [["category_id", 1], ["created_at", Sat, 25 Jan 2014 08:28:43 UTC +00:00], ["image", "file.jpg"], ["slug", "1"], ["updated_at", Sat, 25 Jan 2014 08:28:43 UTC +00:00]]
SQL (1.8ms)  INSERT INTO "artwork_translations" ("artwork_id", "created_at", "locale", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["artwork_id", 100], ["created_at", Sat, 25 Jan 2014 08:28:43 UTC +00:00], ["locale", "en"], ["updated_at", Sat, 25 Jan 2014 08:28:43 UTC +00:00]]
PG::NotNullViolation: ERROR:  null value in column "title" violates not-null constraint
(0.2ms)  ROLLBACK
```

---

I understand that a reason behind it is that `accepts_nested_attributes_for :translations` sets `autosave` association option to **true** and it forces activerecord to save empty associations when I'm saving parent record.
I figured that globalize uses `after_create` callbacks with `autosave: false` by default and `accepts_nested_attributes_for` breaks the save logic, but I need to add/edit multiple translations at once (hence `accepts_nested_attributes_for`).

---

I came up with a workaround to reset `autosave` option, when I'm creating records using generated methods, instead of `translations` association.
But it doesn't seem right and I don't want to put it into production.

``` ruby
[...]

PR #341, opened 2014-03-05
Replace after_ callbacks with before_

scarfacedeb:
Beginning of discussion: https://github.com/globalize/globalize/issues/325

I replaced after_create/after_update with before_create/before_update and set autosave: true.
Now we don't need to explicitly call translations.save inside save_translations!, because autosave takes care of it for us.
Test results haven't changed (192 tests, 1 skips).

Also I added a couple of tests for the current null: false case.

I merged my current branch into the previously created before_callbacks branch and I'm not sure that it would give the best git log. Can anybody validate it for me? (I can fork globalize again and reapply all changes by hand, if it's needed)

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -75,10 +75,10 @@ module Globalize
                                 :foreign_key => options[:foreign_key],
                                 :dependent   => :destroy,
                                 :extend      => HasManyExtensions,
-                                :autosave    => false
+                                :autosave    => true
 
-        after_create :save_translations!
-        after_update :save_translations!
+        before_create :save_translations!
+        before_update :save_translations!
       end
     end
 
--- a/lib/globalize/active_record/adapter.rb
+++ b/lib/globalize/active_record/adapter.rb
@@ -38,8 +38,6 @@ module Globalize
           translation = record.translations_by_locale[locale] ||
                         record.translations.build(locale: locale.to_s)
           attrs.each { |name, value| translation[name] = value }
-          ensure_foreign_key_for(translation)
-          translation.save!
         end
 
         reset
@@ -50,12 +48,6 @@ module Globalize
       end
 
       protected
-
-      # Sometimes the translation is initialised before a foreign key can be set.
-      def ensure_foreign_key_for(translation)
-        # AR >= 4.1 reflections renamed to _reflections
-        translation[translation.class.reflections[:globalized_model].foreign_key] = record.id
-      end
 
       def type_cast(name, value)
         return value.presence unless column = column_for_attribute(name)
--- a/test/data/models/artwork.rb
+++ b/test/data/models/artwork.rb
@@ -1,0 +1,4 @@
+class Artwork < ActiveRecord::Base
+  translates :title
+  accepts_nested_attributes_for :translations
+end
--- a/test/data/schema.rb
+++ b/test/data/schema.rb
@@ -221,4 +221,13 @@ ActiveRecord::Schema.define do
     t.references :post
     t.string :file_type
   end
+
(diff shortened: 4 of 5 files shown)


parndt:
Well I suppose this would go in version 5.1.0 instead, now.

jonasheinrich:
Thanks for the fix and for the merge! 
@parndt Any chance this gets released soon? As far as I can see `5.0.1` doesn't include this fix. Thanks!

parndt:
@jonasheinrich oh, that's a shame.  If this is in the master branch then I suppose it'll get released when 5.1.0 does.  I don't have any planned date for that, but as 5.0.1 is over a year old, it'd make sense!

@globalize/collaborators are we okay with releasing [this list of changes (compare view)](https://github.com/globalize/globalize/compare/v5.0.1...master) as 5.1.0?

tomash:
Yes, it looks like a release with some deep architectural changes (callbacks!) that should be mostly compatible in terms of public api with 5.0.

jonasheinrich:
@tomash @parndt Thanks for your replies! 
Just as a remark: The current head of the master branch somehow freezes my tests (probably some gem incompatibilities). So if you could release just that merge commit or a minimum of following commits that would be great. Thanks in advance!

parndt:
I don't suppose there's any way we can run your tests? :)

jonasheinrich:
Yeah, unfortunately there is no way :D. Which makes it kind of hard for you to guess the correct commit to release. So I'd say just release `40a909b14a11d677efcd2f35b0cbb6abe732cfe8`. But that is only my opinion :)

parndt:
Could you use rebase against the globalize repository to find the commit in which they start freezing?  

You can clone globalize:

```
git clone https://github.com/globalize/globalize
``` 

Then, specify globalize as a path source:

```ruby
gem "globalize", path: "/path/to/your/local/clone"
```

Then, you can run git bisect inside the globalize directory:

```
git bisect bad master
git bisect good v5.0.1
```

And git will switch between commits, allowing you to run your test suite each time it does it and confirm whether or not it freezes.  Then you can tell bisect whether it's a "good" commit or a "bad commit".  Eventually you should track down the right commit 😄

jonasheinrich:
@parndt Thanks for taking so much time. Yes, that's an idea I could have come up with myself. Hopefully gonna do that today.

jonasheinrich:
@parndt Seems like the first bad commit is `c0147df067171df8865fba693fd4521219392634`. Not sure how that is possible looking at this commit. But if you could release the parent of this one (`98ab1aae7ee1af2960cb4e02a15ed80a657b79e4`, right?) it would be neat!

parndt:
I'd rather find out why that makes your tests freeze, as it's an improvement. 
Which test freezes? 
In what way does it freeze?
Which stage of testing freezes?

jonasheinrich:
@parndt I fixed the application (which was not the fault, not the tests). Thanks for being so persuasive :)
Just for your information: The problem was a missing `include(:translations)` such that I got an SQL error because I order by the translated name. 
So you could as well release the `HEAD` of `master`. Thanks again!

jonasheinrich:
@parndt Any chance you can release this? Or is there anything more I can do? Thanks in advance!

parndt:
@globalize/collaborators any objection to a release?

simi:
I think Rails 5 is broken. I try to take a look what happened. Will be nice to release with green CI builds.

bricesanchez:
I'm not a collaborator for globalize but rails 5 update of Refinery needs this PR merged : #505

parndt:
That PR has a checklist which is still not showing as complete, does that sound correct to you?

bricesanchez:
Oh i didn't see that #501 was merged, sounds correct for me.

PR #348, opened 2014-05-18
added inverse_of to the relationships

mkon:
Hi everyone,

I added the :inverse_of to the relationships between the translations and the globalized_model. This way you can for example access the globalized_model from the translations even before the translated model is saved.

I for example have a model like this

``` ruby
class Page

  translates :path, :title
  accepts_nested_attributes_for :translations

  class Translation < Globalize::ActiveRecord::Translation

    before_validation :calculate_strings

    def calculate_strings
      # do something depending on globalized_model
    end
  end
end
```

which will currently not work when saving new pages.

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -74,7 +74,8 @@ module Globalize
                                 :foreign_key => options[:foreign_key],
                                 :dependent   => :destroy,
                                 :extend      => HasManyExtensions,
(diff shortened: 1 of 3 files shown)


mkon:
any news on this? :) Rebased it to master.

mkon:
ping

everplays:
Can we get this merged? I don't wanna monkey patch Globalize for something like this.

PR #349, opened 2014-05-20
Fixed Carrierwave(probably other gems too) integration issue

tomasz-jasiulek:
Hey guys,

Recently I have tried to integrate carrierwave with globalize and faced a problem when trying to translate fields with mounted Carrierwave Uploader. The uploader is not triggered when saving the object, so I have decided to look deeper into globalize gem. It looks like AR Adapter is operating on too low api level when assigning translated values. Assigining via

``` ruby
attrs.each { |name, value| translation[name] = value }
```

skips Carrierwave(and probably other gems) uploader behavior. I have written fix for this, which uses accessor methods for assignment rather than low level hash assignment. Now everything works fine and do not need any hacking around.

### DIFF
--- a/lib/globalize/active_record/adapter.rb
+++ b/lib/globalize/active_record/adapter.rb
@@ -37,7 +37,10 @@ module Globalize
         stash.reject {|locale, attrs| attrs.empty?}.each do |locale, attrs|
           translation = record.translations_by_locale[locale] ||
                         record.translations.build(locale: locale.to_s)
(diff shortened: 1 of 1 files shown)


partydrone:
Has there been any movement on this issue? I'm running into problems because of this.

parndt:
@partydrone did applying the patch work for you?

partydrone:
No. I created a new issue #456 describing the problem.

parndt:
Thanks!

PR #352, opened 2014-05-29
Fixes a problem with dup and instance variables

hypomodern:
Fixes a problem where after dup the dup'd model and the original model share a translation instance, which means that if you mutate a translated field on the dup and save it, the original becomes a clone of the dup.

We were running in to this in our efforts to use [spree](https://github.com/spree/spree_i18n), so there's at least one major implementation that is effected.

Thanks in advance!


nskeip:
@hypomodern Thank you very much!

PR #356, opened 2014-06-11
Add UUID support

sebastiankliem:
I've added support for using UUIDs in the translation tables when the model that should be translated uses UUIDs instead of IDs itself. Therefore I have extended create_translation_table with an option "use_uuid".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ NOTES
 .DS_Store
 .bundle
 Gemfile.lock
-globalize.iml
 .rvmrc
 *.sw*
 *.gem
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ NOTES
 .DS_Store
 .bundle
 Gemfile.lock
+globalize.iml
 .rvmrc
 *.sw*
 *.gem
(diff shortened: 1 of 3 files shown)


parndt:
Configurable would be super nice

sebastiankliem:
@parndt Unfortunately I do not have the time to work on this any further. I'm sorry for that.

parndt:
@sebastiankliem that's fine, thank you so much for all of the work you have done!

daniel-de-wit:
Just and idea, maybe we can check the delegated model for what kind of type it uses?

```
connection.create_table(translations_table_name, id: model.columns_hash[model.primary_key].type.to_sym)
```

I know this is ugly, i'm still kind of new to all this.

gregmolnar:
I think @daniel-de-wit 's solution could be a good approach just the test suite needs to be run against postgres as well. In the meantime I think the simples solution is to alter the column in the migration:

```
Post.create_translation_table! :title => :string, :text => :text
change_column :post_translations, :post_id, :uuid
```

abhiQmar:
This should be closed as UUID is now being supported:
[link](https://github.com/globalize/globalize/blob/master/lib/globalize/active_record/migration.rb#L73)

Issue #372, opened 2014-10-07
dirty tracking errors


momolog:
The following tests (I added to test/globalize/dirty_tracking_test.rb):

``` ruby
    it 'works for assigning new value == old value of other locale' do
      post = Post.create(:title => nil, :content => 'content')
      assert_equal [], post.changed

      post.title = 'english title'
      assert_equal ['content', 'title'], post.changed

      I18n.locale = :de
      post.title  = nil
      assert_equal ['content', 'title'], post.changed
    end
```

fails for the first assertion with

```
   Expected: []
      Actual: ["content"]
```

and for the 3rd with 

```
    Expected: ["content", "title"]
      Actual: ["content"]
```

Imho. these assumptions should all pass. These are bugs, right?

This can cause ugly, hard-to-track bugs in situations like the following:
1) all translations of `object`are empty
2) an update sets the english translation to a value and the german to nil right after
3) `object.changed?` yields `false` though the english value has clearly changed

I will prepare a PR for this.

PR #373, opened 2014-10-07
keep object changed when re-setting to old value

momolog:
this solves the problem that the value can be set for a *different
locale* so it does not really re-set the value for the original locale
(#372)

a real solution would keep track of the old values *per locale*


parndt:
Closing in favour of #379 - thanks!

PR #380, opened 2014-11-01
attempt to make it work with rails 4.2

ncri:
I had to comment dirty tracking for Rails 4.2 for now because it causes rails trying to save the translated attributes on the untranslated object. It looks like rails changed the way it saves models, using the changed attributes array to determine the columns to be saved.


leighhalliday:
@ncri with the new code for working with Rails 4.2, if I have a `title` field in both my main `posts` table and also my `post_translations` table, it isn't being set. Is this either an error on my part (I shouldn't keep that column in the main posts table), or is this an issue with a change needed for Rails 4.2. Before I upgraded to Rails 4.2 it didn't seem to be an issue.

parndt:
Now that Rails 4.2 is released, I'm keen to push hard to get this merged to master.

ncri:
@leighhalliday: I'm not sure how globalize should behave in this case. If it was setting the field in the main table before rails 4.2 it was probably the defined behavior. Although I don't quite see the use case. Why would you keep the field in the main table? But maybe I'm missing something here.

@parndt: unfortunately I don't have much time to work on this right now. But I think if someone can help a bit with investigating and fixing the failing tests, we shouldn't be too far of getting it working. However, I'm not sure if my patching approach is future proof enough or if deeper architectural changes in globalize are needed. My branch works fine in a production app, although it doesn't heavily rely on globalization.

parndt:
I've created a branch, [rails-4-2-upgrade](https://github.com/globalize/globalize/compare/rails-4-2-upgrade?expand=1)

PR #389, opened 2014-12-08
Add `order` method to globalise.

qnm:
Populating `order` with a translated column name does not currently result in automatic mapping to the translation table.

I've added an `order` method to `ActiveRecord::QueryMethods`, which will map a translated column to the translation table when called with either `order(:title)` or `order(title: :desc)` format.

### DIFF
--- a/lib/globalize/active_record/query_methods.rb
+++ b/lib/globalize/active_record/query_methods.rb
@@ -22,6 +22,14 @@ module Globalize
         end
       end
 
+      def order(opts, *rest)
+        if respond_to?(:translated_attribute_names) && parsed = parse_translated_order(opts)
+          super(parsed)
+        else
+          super
+        end
+      end
+
       def exists?(conditions = :none)
         if parsed = parse_translated_conditions(conditions)
           with_translations_in_fallbacks.exists?(parsed)
@@ -62,6 +70,27 @@ module Globalize
         else
           relation.with_translations_in_fallbacks
         end
+      end
+
+      private
+
+      def parse_translated_order(opts)
+        case opts
+        when Hash
+          ordering = opts.map do |column, direction|
+            klass = translated_column?(column) ? translation_class : self
+            klass.arel_table[column].send(direction)
+          end
+          order(ordering).order_clauses
+        when Symbol
+          translated_column_name(opts) if translated_attribute_names.include?(opts)
+        else # failsafe returns nothing
+          nil
+        end
+      end
+
+      def translated_column?(column)
+        translated_attribute_names.include?(column)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/globalize/active_record/query_methods.rb
```diff
@@ -63,6 +71,25 @@ def join_translations(relation = self)
...
+        case opts
+        when Hash
+          ordering = opts.map do |column, direction|
+            klass = translated_column?(column) ? translation_class : self
+            klass.arel_table[column].send(direction)
+          end
+          order(ordering).order_clauses
+        when Symbol
+          translated_column_name(opts) if translated_attribute_names.include?(opts)
+        end
```
parndt: I believe it would be prudent to ensure that the case statement has a default case

qnm:
@parndt I've added an explicit nil return to the `parse_translated_order` method, as any args other then `Hash` and `Symbol` aren't parseable. The existing tests should cover this case already.

Shout out if that's not what you had in mind!

lib/globalize/active_record/query_methods.rb
```diff
@@ -22,6 +22,16 @@ def where(opts = :chain, *rest)
...
+      def order(opts, *rest)
+        super unless respond_to?(:translated_attribute_names)
+
+        if parsed = parse_translated_order(opts)
+          super(parsed)
+        else
+          super
+        end
+      end
+
```
parndt: well, it calls super twice, but with different arguments..

parndt:
@qnm seems to be failing CI in rubinius.. can you determine whether this is one of your changes, or whether you just need to rebase against the current version of master? (it passes in master)

qnm:
@parndt Are you able to re-run that build? The error seems Travis-specific:

```
It seems your ruby installation is missing psych (for YAML output).

To eliminate this warning, please install libyaml and reinstall your ruby.

An exception occurred running /home/travis/.rvm/gems/rbx-2.4.1@global/gems/rake-10.4.2/lib/rake/rake_test_loader.rb:

no such file to load -- psych (LoadError
```

parndt:
@qnm doesn't seem to have helped.. hmm.

parndt:
rbx-2 is just failing lately; I've marked it as allowed failure on CI for now.

parndt:
@qnm thanks for this change; are you able to squash this down into a single or a couple commits so that people can more easily use `git blame` in future please? :heart:

qnm:
@parndt I've squashed to one; most of the other commits are refactoring. Thanks!

parndt:
:heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart::heart:

PR #396, opened 2015-01-05
Rails 4.2 upgrade

parndt:
Not yet ready to merge, current status: 

```
195 tests, 492 assertions, 3 failures, 2 errors, 1 skips
```

Carrying on from #380

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,9 @@ script: "bundle exec rake"
 install:
   - bundle install --retry=3
 env:
-  - RAILS_4_2=true DB=sqlite3
-  - RAILS_4_2=true DB=postgres
-  - RAILS_4_2=true DB=mysql
-  - RAILS_4_1=true DB=sqlite3
-  - RAILS_4_1=true DB=postgres
-  - RAILS_4_1=true DB=mysql
-  - RAILS_4_0=true DB=sqlite3
-  - RAILS_4_0=true DB=postgres
-  - RAILS_4_0=true DB=mysql
+  - DB=sqlite3
+  - DB=postgres
+  - DB=mysql
 rvm:
   - 2.2
   - 2.1
--- a/globalize.gemspec
+++ b/globalize.gemspec
@@ -15,10 +15,10 @@ Gem::Specification.new do |s|
   s.require_path = 'lib'
   s.rubyforge_project = '[none]'
 
-  s.add_dependency 'activerecord', '>= 4.2.0.beta4', '< 4.3'
-  s.add_dependency 'activemodel', '>= 4.2.0.beta4', '< 4.3'
+  s.add_dependency 'activerecord', '>= 4.2.0', '< 4.3'
+  s.add_dependency 'activemodel', '>= 4.2.0', '< 4.3'
 
-  s.add_development_dependency 'database_cleaner', '~> 1.3.0'
+  s.add_development_dependency 'database_cleaner', '~> 1.4.0'
   s.add_development_dependency 'minitest'
   s.add_development_dependency 'minitest-reporters'
 
--- a/globalize.gemspec
+++ b/globalize.gemspec
@@ -15,19 +15,8 @@ Gem::Specification.new do |s|
   s.require_path = 'lib'
   s.rubyforge_project = '[none]'
 
-  if ENV['RAILS_4_0']
-    s.add_dependency 'activerecord', '~> 4.0.0'
-    s.add_dependency 'activemodel', '~> 4.0.0'
-  elsif ENV['RAILS_4_1']
-    s.add_dependency 'activerecord', '~> 4.1.0'
-    s.add_dependency 'activemodel', '~> 4.1.0'
-  elsif ENV['RAILS_4_2']
-    s.add_dependency 'activerecord', '>= 4.2.0.beta2', '< 4.3'
-    s.add_dependency 'activemodel', '>= 4.2.0.beta2', '< 4.3'
-  else
-    s.add_dependency 'activerecord', '>= 4.0.0', '< 5'
-    s.add_dependency 'activemodel', '>= 4.0.0', '< 5'
-  end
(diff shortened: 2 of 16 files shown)


parndt:
@globalize/collaborators @globalize/owners @ncri  I'd like to change it so that when writing, the original column is completely ignored.  Currently the column is still updated if writing in the default locale but this hurts us in terms of failing tests and also by way of crufty code that we need for this. 

I don't think it makes sense (and never have) to translate a column but keep writing to the original table.  This would mean that if you ever changed the default locale, you'd lose a lot of information if you relied on the original table.

Thoughts?

bricesanchez:
@parndt You're right, i think it's useless to keep writing to the original table. If someone decide to remove globalize after a long time, he should migrate the content he want to keep in the original table.

ncri:
@parndt Totally agree, I was confused when I noticed that globalize does this. Wonder if anyone has a usecase for this?

parndt:
Yes people often mention this use case but I think it would be easy enough to write an extension to do it instead, I don't think we should do it by default.

parndt:
I've squashed commits and rebased against master. This should make the branch much easier to work with as well as removing lots of deprecations warnings.

parndt:
@qnm hey your patch for ordering (#389) doesn't work for Rails >= 4.2:

```
ERROR["test_0002_returns record in order, column and direction as hash", #<Class:0x007fce0c2cad58>, 1.643924]
 test_0002_returns record in order, column and direction as hash#finder methods::.order::returns record in order::for translated columns (1.64s)
NoMethodError:         NoMethodError: undefined method `order_clauses' for #<ActiveRecord::Relation []>
            ~/.gem/ruby/2.2.0/gems/activerecord-4.2.0/lib/active_record/relation/delegation.rb:136:in `method_missing'
            ~/.gem/ruby/2.2.0/gems/activerecord-4.2.0/lib/active_record/relation/delegation.rb:99:in `method_missing'
            globalize/lib/globalize/active_record/query_methods.rb:85:in `parse_translated_order'
            globalize/lib/globalize/active_record/query_methods.rb:26:in `order'
            globalize/test/globalize/translated_attributes_query_test.rb:192:in `block (5 levels) in <class:TranslatedAttributesQueryTest>'
        ~/.gem/ruby/2.2.0/gems/activerecord-4.2.0/lib/active_record/relation/delegation.rb:136:in `method_missing'
        ~/.gem/ruby/2.2.0/gems/activerecord-4.2.0/lib/active_record/relation/delegation.rb:99:in `method_missing'
        globalize/lib/globalize/active_record/query_methods.rb:85:in `parse_translated_order'
        globalize/lib/globalize/active_record/query_methods.rb:26:in `order'
        globalize/test/globalize/translated_attributes_query_test.rb:192:in `block (5 levels) in <class:TranslatedAttributesQueryTest>'
```

qnm:
@parndt Thanks - I'll get look into that tomorrow.

qnm:
I've dropped some fixes for translated ordering into #403

parndt:
current status:

```
Started with run options --seed 17425

 SKIP["test_0001_does not update original columns with content not in the default locale", #<Class:0x007fb445e21e80>, 1.160004]
 test_0001_does not update original columns with content not in the default locale#model with attribute column in both model and translation table (1.16s)
        need to fix before we release 4.0.0
        /code/parndt/globalize/test/globalize/attributes_test.rb:211:in `block (2 levels) in <class:AttributesTest>'

 FAIL["test_0002_updates original columns with content in the default locale", #<Class:0x007fb445e21e80>, 1.170458]
 test_0002_updates original columns with content in the default locale#model with attribute column in both model and translation table (1.17s)
        Expected: "New Title"
          Actual: nil
        /code/parndt/globalize/test/globalize/attributes_test.rb:231:in `block (2 levels) in <class:AttributesTest>'

 FAIL["test_0004_updates original columns with content in the same locale", #<Class:0x007fb445e21e80>, 1.189039]
 test_0004_updates original columns with content in the same locale#model with attribute column in both model and translation table (1.19s)
        Expected: "ontvriend"
          Actual: nil
        /code/parndt/globalize/test/globalize/attributes_test.rb:276:in `block (2 levels) in <class:AttributesTest>'

 FAIL["test_0003_does not update original columns with content in a different locale", #<Class:0x007fb445e21e80>, 1.195365]
 test_0003_does not update original columns with content in a different locale#model with attribute column in both model and translation table (1.20s)
        Expected: "ontvrienden"
          Actual: nil
        /code/parndt/globalize/test/globalize/attributes_test.rb:246:in `block (2 levels) in <class:AttributesTest>'

  204/204: [=================================================================================================================================] 100% Time: 00:00:01, Time: 00:00:01

Finished in 1.72506s
204 tests, 506 assertions, 3 failures, 0 errors, 1 skips
```

I'm inclined to just delete these failing tests as we don't want to support this use case anymore.  I think we should instead update our documentation to state that this is not supported and that it may or may not work but is an untested case.

mvdstam:
@parndt Agreed. When removing Globalize, people should just use the migrate_data option in their migrations. From my point of view, Globalize writing stuff for translated columns to the original table should be considered unexpected behavior. You could just invert the tests rather then removing them altogether, since you wouldn't want this behavior anyway.

parndt:
Cool! Well, this seems like it'd be ready to be merged to master then?

mvdstam:
I guess so!

Perhaps bump the version number as well and push it to ruby gems as well? Since this is kind of a big thing, plus friendly_id-globalize can then just depend on that newest Globalize version instead of this branch it's now pointing to and have its version number bumped as well.. Nice and fresh gems for rails 4.2; I'm loving it.

parndt:
Cool, it's not backwards compatible so it would have to be a new major version, I think?

mvdstam:
Yes, I'd bump it to 5.0.0.

parndt:
This has been released as version 5.0.0

If there are other things to do, we can just use patch releases.

PR #398, opened 2015-01-13
fix issue with initializing new model in relation using has_many :through which has no translations

Azdaroth:


Azdaroth:
There was a problem when building new model with ``` has_many :through ``` when intermediate model is translatable, but the model being built is not (it raised ``` undefined local variable or method `translations_table_name' ```). That PR is supposed to fix this.

parndt:
Looks good to me.

Issue #392, opened 2015-01-03
Serialized attributes deprecated in R4.2, due for removal in R5


JohnSmall:
Creating a new project in Rails 4.2 I get tons of deprecation warnings about serialized attributes being slated for removal in R5. It's obviously an issue that will have to be addressed. 

The PaperTrail guys have the same issue see here https://github.com/airblade/paper_trail/issues/416 . The thread includes code for an initializer to silence the warnings and I've modified it for globalize ;-

    current_behavior = ActiveSupport::Deprecation.behavior
    ActiveSupport::Deprecation.behavior = lambda do |message, callstack|
      return if message =~ /`serialized_attributes` is deprecated without replacement/ && callstack.any? { |m| m =~ /paper_trail|globalize/ }
      Array.wrap(current_behavior).each { |behavior| behavior.call(message, callstack) }
    end

parndt:
Seems like we should be doing something [like this](https://github.com/rails/rails/pull/15704/files#diff-a72ce03f1fc846ef77d68506c9060bfcR62)

parndt:
And possible something [like this](https://github.com/rails/rails/pull/15704/files#diff-d5fc467bd1cb0f9a9b1f8f2dec3ab8c8R212)

simi:
fixed by #402, #407 and #412

PR #402, opened 2015-01-25
Get serializer for attribute without `serialized_attributes`

nburkley:
* 'serialized_attributes` is deprecated without replacement in Rails 5.
* Use `column.cast_type.coder` instead

Addresses issue https://github.com/globalize/globalize/issues/392


parndt:
Thank you!

PR #403, opened 2015-01-30
Fix ordering via translated columns for rails-4-2

qnm:
@parndt I've updated the ordering code and it _appears_ to work correctly. I don't have a rails-4-2 app with globalize to test for reals, but it the ordering tests pass. Feel free to ping me if you need any further assistance!

This relates to PR #396


parndt:
Thank you!

PR #404, opened 2015-01-31
Fixed correct table name for "news translations"

mvdstam:
This fixes the error that occurs for test "create a translation class for model with custom table name".


parndt:
Thank you!

PR #405, opened 2015-02-02
Inverted tests for updating original columns in some cases.

mvdstam:
All tests now pass. The tests themselves are renamed accordingly.


mvdstam:
#396

PR #407, opened 2015-02-05
Initialize Globalize just from model metadata.

simi:
In refinery we found a serious bug.

When you try to run clean migrations, the DB table is not present yet. But `serializer_for_attribute` is using methods that needs table meta data and that cause crash you can see on [RefineryCMS CI](https://travis-ci.org/refinery/refinerycms/jobs/49285383). This was introduced in https://github.com/globalize/globalize/pull/402.

Since `serialized_attributes` are deprecated, we need to use similar approach that don't need actual table but can use model metadata. This is really naive and bad implementation, but it works.

Alternatively, we can revert #402 and get some time until Rails 5 lands (a lot of time now).

@parndt @shioyama

### DIFF
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -1,6 +1,7 @@
 require 'active_record'
 require 'patches/active_record/xml_attribute_serializer'
 require 'patches/active_record/query_method'
+require 'patches/active_record/serialization'
 require 'patches/active_record/uniqueness_validator'
 require 'patches/active_record/persistence'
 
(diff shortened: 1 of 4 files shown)


simi:
hmm, actually this wasn't meant to be merged now @parndt

simi:
I think it will be better to revert #402 and open issue how to solve that deprecation warning in classy way.

parndt:
I just wanted to solve the issue so that we could discuss about how to further improve it.

PR #408, opened 2015-02-06
`#cache_key` should change when translation changes

heathd:
Following up from #330, this is an alternate solution. Rather than forcing the 'parent' model object to be updated if the translation changes, we modify the implementation of `#cache_key` for translated models so that it incorporates the cache_key of the translation as well.

This results in a cache key of:

`products/1-20150206003237457906000/product/translations/new`

for a new translation

and 

`products/1-20150206000856438942000/product/translations/1-20150206000856442618000`

for an existing translation

I'd welcome comments and feedback, in particular if @shlensky could try this out to see if it works in his codebase.

### DIFF
--- a/test/data/models/product.rb
+++ b/test/data/models/product.rb
@@ -1,0 +1,3 @@
+class Product < ActiveRecord::Base
+  translates :name
+end
--- a/lib/globalize/active_record/class_methods.rb
+++ b/lib/globalize/active_record/class_methods.rb
@@ -47,7 +47,7 @@ module Globalize
             klass = self.const_set(:Translation, Class.new(Globalize::ActiveRecord::Translation))
           end
 
-          klass.belongs_to :globalized_model, :class_name => self.name, :foreign_key => translation_options[:foreign_key], :touch => true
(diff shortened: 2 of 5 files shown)


parndt:
cc @tekin

shlensky:
:+1:

test/data/models/product.rb
```diff
@@ -0,0 +1,3 @@
+class Product < ActiveRecord::Base
```
simi: Is new model really needed?

heathd: none of the other models had timestamps on. The timestamps are needed to exhibit the fact that the `cache_key` changes on modification of the translation. 

I considered adding timestamps to an existing model but I was concerned about unintended side-effects, so I opted for added a new model instead.

I'll add note about this to the commit message.

heathd:
reworded another one of the commit messages to explain why we removed `:touch => true`

tekin:
I approve, this would make globalize and Rails optimistic locking work together again :+1:

heathd:
:+1: thanks for merging

PR #411, opened 2015-02-10
Remove handling for locale attribute on parent model

tekin:
This removes some legacy code that was added to automatically keep a default translation in sync with columns on the parent table[1]. Globalize will no longer support this feature and users are expected to drop the translated columns from parent table.

[1] globalize#183


parndt:
You're so cool

tekin:
Why thank you ;)

simi:
SICK!!!

shioyama:
great stuff!! Much cleaner. :sparkles:

simi:
*Just note: it will be nice to add tests to check this would not be added again by mistake.*

Issue #390, opened 2014-12-23
Regression - Globalize not anymore thread safe


mtparet:
Because of https://github.com/globalize/globalize/pull/374, now we can't anymore use Globalize in threaded contexts like Sidekiq. (Because now the current locale is shared across all threads instead of being specific to the current thread)

Seems like an important breaking change.

Following test succeeded before the commit removing usage of `Thread.current` :
```ruby
  describe 'with locale' do
    def perform_with_locale(locale)
      Thread.new do
        Globalize.with_locale(locale) do
          sleep 1
          puts "#{Thread.current} : locale #{locale}, fallback: #{Globalize.locale} \n"
          expect(Globalize.locale).to eq(locale)
        end
      end
    end

    it 'succeed in one thread' do
      thread = perform_with_locale(:end)
      thread.join
    end

    it 'succeed multithreading' do
      thread1 = perform_with_locale(:en)

      thread2 = perform_with_locale(:fr)

      thread3 = perform_with_locale(:de)

      thread1.join && thread2.join && thread3.join
    end
  end
```

Dschee:
Would love to see this fixed, it is a really annoying bug in combination with sidekiq! I find #374 to be less important than this and therefore hope that @hube's fix is reverted until someone finds a better solution. 

Globalize is pretty useless for me if I can't update any data in more than one language in the background with sidekiq. @parndt Could you have a look and revert if possible?

simi:
@Dschee @hube :( I haven't problem with `Thread.current`. But maybe some config value would be good here since someone will need to use https://github.com/steveklabnik/request_store for example. Any idea for API?

scarfacedeb:
Fixed in #420

Issue #224, opened 2013-04-16
Globalize.locale remaining set over different sessions


gaganawhad:
When I set `Globalize.locale`, it is remains set even over different sessions when using  webservers like thin and unicorn. I am guessing this is unintentional behavior. 

This seems related to this issue: https://github.com/rails/rails/issues/8517 and whats seems to be recommended is "cleaning up the Thread". 

Is this something that needs to be looked into? Let me know if I can help.

bitboxer:
Any update on this?

gaganawhad:
Hey @bitboxer, sorry I didn't get to this. Not sure that I can anytime soon... :(

PR #413, opened 2015-02-15
Make Globalize thread safe by default again.

simi:
I think it is easy enough to use custom storage just by overriding storage method on Globalize. I was thinking about getter and setter for that, but proc will be needed there probably and it looks ugly in config.

https://github.com/globalize/globalize/commit/37c19b5f8b19ba14bc89f7237064857d46ad815d :red_circle: fails without https://github.com/globalize/globalize/commit/b1857194a147ef2a0d3bd99d3c7cab17d0683edc :green_heart: 

related to #390 #374 

ping @mtparet @hube @parndt @Dschee

### DIFF
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -5,7 +5,6 @@ require 'patches/active_record/serialization'
 require 'patches/active_record/uniqueness_validator'
 require 'patches/active_record/persistence'
 
-
 module Globalize
   autoload :ActiveRecord, 'globalize/active_record'
   autoload :Interpolation,   'globalize/interpolation'
@@ -52,18 +51,22 @@ module Globalize
       i18n_fallbacks? ? I18n.fallbacks[for_locale] : [for_locale.to_sym]
     end
 
+    def storage
+      Thread.current
+    end
+
   protected
 
     def read_locale
-      @globalize_locale
+      storage[:globalize_locale]
     end
 
     def set_locale(locale)
-      @globalize_locale = locale.try(:to_sym)
+      storage[:globalize_locale] = locale.try(:to_sym)
     end
 
     def read_fallbacks
-      @fallbacks || HashWithIndifferentAccess.new
+      storage[:globalize_fallbacks] || HashWithIndifferentAccess.new
     end
 
     def set_fallbacks(locales)
@@ -73,7 +76,7 @@ module Globalize
         fallback_hash[key] = value.presence || [key]
       end if locales.present?
 
-      @fallbacks = fallback_hash
+      storage[:globalize_fallbacks] = fallback_hash
     end
   end
 end
--- a/test/globalize_test.rb
+++ b/test/globalize_test.rb
@@ -217,4 +217,28 @@ class GlobalizeTest < MiniTest::Spec
       end
     end
   end
+
+  describe 'with locale' do
+    def perform_with_locale(locale)
+      Thread.new do
+        Globalize.with_locale(locale) do
+          sleep 1
+          assert_equal locale, Globalize.locale
+        end
+      end
+    end
+
+    it 'succeed in one thread' do
+      thread = perform_with_locale(:end)
+      thread.join
+    end
+
+    it 'succeed multithreading' do
+      thread1 = perform_with_locale(:en)
+      thread2 = perform_with_locale(:fr)
+      thread3 = perform_with_locale(:de)
+
+      thread1.join && thread2.join && thread3.join
+    end
+  end
 end


lib/globalize.rb
```diff
@@ -73,7 +76,7 @@ def set_fallbacks(locales)
         fallback_hash[key] = value.presence || [key]
       end if locales.present?
 
-      @fallbacks = fallback_hash
+      Thread.current[:globalize_fallbacks] = fallback_hash
```
parndt: This can't use storage?

simi: sec

simi: good catch @parndt, fixed

mtparet:
I didn't take the time for this, thank you @simi ! :heart:

test/globalize_test.rb
```diff
@@ -217,4 +217,28 @@ class GlobalizeTest < MiniTest::Spec
       end
     end
   end
+
+  describe 'with locale' do
+    def perform_with_locale(locale)
+      Thread.new do
+        Globalize.with_locale(locale) do
+          sleep 1
```
parndt: Oh no, this will make our tests ~75% slower! :wink:

mtparet: 0.1 should be enough :)

simi: Actually not. I tested it.
Dne 16.2.2015 1:22 "Matthieu Paret" <notifications@github.com> napsal(a):

mtparet: :-( it depends of the system.

simi: I have really poor one.
Dne 16.2.2015 1:41 "Matthieu Paret" <notifications@github.com> napsal(a):

scarfacedeb:
Maybe, we should use [request_store](https://github.com/steveklabnik/request_store), instead of Thread.current to prevent an unwanted sharing of values in multiple requests?
I'm not a thread expert, but it seems sane.

simi:
@scarfacedeb RequestStore is meant to be used in Rack application as Rack middleware. So it is not usable in Sidekiq for example.

scarfacedeb:
@simi OK. 
I've found an old issue (https://github.com/globalize/globalize/issues/224) about that kind of unwanted behaviour and @parndt also suggested to use request_store.

parndt:
@steveklabnik hey, would you recommend using request_store for this?

steveklabnik:
Yes, this code is effectively RequestStore.

@simi while it's true that I concieved it as being out of Rack, there's no reason why you couldn't use it in Sidekiq. You'd just need to make sure to call the clear function at the right time. I don't use Sidekiq, so I don't know what kind of hooks it has, off the top of my head.

parndt:
Thanks, @steveklabnik !

@mperham do you have any wisdom here? Would it be conceiveable that this would work with Sidekiq?

steveklabnik:
https://github.com/steveklabnik/request_store/blob/master/lib/request_store/middleware.rb#L7-L11 <- you'd need to call that method (`clear!`) on startup. given https://github.com/mperham/sidekiq/wiki/Middleware seems fine.

I would not mind having said middleware inside of the gem itself, if that's helpful.

mperham:
Here's Sidekiq's i18n middleware: https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/middleware/i18n.rb  Yeah, `Thread.current` is totally cool with Sidekiq; I use it everywhere.

simi:
This is not just about Sidekiq, but about non-rack env globally.

What about to stick with `Thread.current` and add `Globalize.clear!` method to allow manual clear when you'll need it. For example in Rails controller, you can use it in `after_filter`.

parndt:
Thanks, @mperham !

scarfacedeb:
@simi Wouldn't it basically duplicate RequestStore?
What if we use RequestStore which would provide automatic clearing on every request.
And non-Rack apps could call RequestStore.clear!

That way the most popular case (in a rails app) is getting better thread safety for free.

simi:
That makes sense @scarfacedeb. Would you mind to send PR for that?

scarfacedeb:
@simi Sure! I'll look into it.

test/globalize_test.rb
scarfacedeb: @simi Could you clarify why is `sleep` needed here?

Is it to give a scheduler time to switch into another thread?
Maybe we could use `Thread.pass` instead

simi: Just copy paste from https://github.com/globalize/globalize/issues/390#issue-52729253. Feel free to make red test (revert my patch - use ivars as storage) and then make it green with `RequestStore`.

scarfacedeb: This test is green with or without `sleep` on my machine. That's why I thought that maybe I'm missing something.
I'll use it without `sleep` then, until there's a reason to have it there.

simi: Try to run it with sleep and ivars storage.

scarfacedeb: I see what you mean.
With ivar and sleep: it fails, but when I remove the sleep - it suddenly passes.
The best news is that it also fails with `Thread.pass` and ivars, but passes with Thread.current.
I suppose it means that we can replace sleep with Thread.pass :)

simi: Good catch! Let's use `Thread.pass`.

simi:
#420 done it.

Issue #391, opened 2015-01-02
Eager loading fails on STI


sbuhr:
Is there any way to eager load the translations? I always end up in n+1 queries.

`Model.includes(:translations)` doesn't work. This occurs in a STI table only.

Any help appreciated.

Edit: STI is supported very well. In my case the AASM gem was outdated.

parndt:
@sbuhr does `Model.joins(:translations)` work? That's what [Refinery CMS is currently using], resulting in a single query.(https://github.com/refinery/refinerycms/blob/master/pages/lib/refinery/pages/finder.rb#L32-L41)

sbuhr:
@parndt No, it doesn't. As mentioned it occurs in STI tables only.

```
Sport.joins(:translations)
  Sport Load (10.8ms)  SELECT `sport_entities`.* FROM `sport_entities` INNER JOIN `sport_entity_translations` ON `sport_entity_translations`.`sport_entity_id` = `sport_entities`.`id` WHERE `sport_entities`.`type` IN ('Sport')
  SportEntity::Translation Load (0.7ms)  SELECT `sport_entity_translations`.* FROM `sport_entity_translations`  WHERE `sport_entity_translations`.`sport_entity_id` = 1519
  SportEntity::Translation Load (0.7ms)  SELECT `sport_entity_translations`.* FROM `sport_entity_translations`  WHERE `sport_entity_translations`.`sport_entity_id` = 1519
  SportEntity::Translation Load (0.6ms)  SELECT `sport_entity_translations`.* FROM `sport_entity_translations`  WHERE `sport_entity_translations`.`sport_entity_id` = 1519
  SportEntity::Translation Load (0.7ms)  SELECT `sport_entity_translations`.* FROM `sport_entity_translations`  WHERE `sport_entity_translations`.`sport_entity_id` = 1520
  SportEntity::Translation Load (0.5ms)  SELECT `sport_entity_translations`.* FROM `sport_entity_translations`  WHERE `sport_entity_translations`.`sport_entity_id` = 1520`
...
```

It loads all 3 translations in single queries.

parndt:
So to clarify, you're saying it eager loads properly with STI tables or it fails to eager load *only* on STI?

sbuhr:
It fails to eager load only on STI.


> Am 05.01.2015 um 22:25 schrieb Philip Arndt <notifications@github.com>:
> 
> it fails to eager load only on STI?
>

parndt:
[...]

PR #415, opened 2015-02-16
Add issue template.

simi:
It will be nice to point to this template in [CONTRIBUTING.md](https://github.com/globalize/globalize/blob/master/CONTRIBUTING.md). Can anyone provide some good copy for that?

### DIFF
--- a/issue_template.rb
+++ b/issue_template.rb
@@ -1,0 +1,38 @@
+# Activate the gem you are reporting the issue against.
+gem 'activerecord', '4.2.0'
+gem 'globalize', '5.0.1'
+require 'active_record'
+require 'globalize'
+require 'minitest/autorun'
+require 'logger'
+
+# Ensure backward compatibility with Minitest 4
+Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)
+
+# This connection will do for database-independent bug reports.
+ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
+ActiveRecord::Base.logger = Logger.new(STDOUT)
+
+ActiveRecord::Schema.define do
+  create_table :posts, force: true do |t|
+  end
+
+  create_table :post_translations, force: true do |t|
+    t.references :post
+    t.string     :title
+    t.text       :content
+    t.string     :locale
+  end
+end
+
+class Post < ActiveRecord::Base
+  translates :content, :title
+end
+
+class BugTest < Minitest::Test
+  def test_association_stuff
+    post = Post.create!(title: 'HI')
+
+    assert_equal 'HI', post.title
+  end
+end


parndt:
What's this do?

simi:
Basic template how to reproduce an issue. It will be nice to have script to reproduce issue and this can make it easier to create one IMHO (actually it is stolen from rails and modified).

simi:
@shioyama @scarfacedeb any ideas? Do you think this is useful at all?

scarfacedeb:
I like the idea.

But it's often the particular configuration of someone's app and models, or an interaction of globalize and other gems that results in an issue.
And this kind of issues would probably disappear when you try to reproduce them using your template.

With that in mind, I think it's worth mentioning in the CONTRIBUTING.md or in a wiki the basic steps that could help to pinpoint a source of some issues.

For example, for the issues with other gems we can advise to remove any other macros from the model one by one. (e.g. https://github.com/globalize/globalize/issues/391)

Also, we could mention that it's better to show the model and schema.rb files in the issue right away.
It would allow to skip the first question that often asks to show those files.

simi:
That's the spirit. This is minimal template. You should copy your behavior to this template piece by piece to get your test case failing. You'll get schema and model also.

According to #391 the process will be similar, but just reversed. You'll be copying behavior one by one to our template. Once you'll have a minimal script to reproduce, you can rethink if that's really a globalize issue.

This idea is stolen from Rails where this works very well.

scarfacedeb:
Cool! I didn't know that Rails is also using it.

I suppose this minimal template isn't the final result, because it won't run as is.
How do you plan to make it run out of the box - a little rails app or more stripped down version with a couple of files (e.g. Gemfile, reproduce.rb, models.rb, etc)?

simi:
Actually it runs out of the box here. What problems do you have?

scarfacedeb:
It was only my own stupidity :)
It runs out of the box now.

I think this minimal template is enough to give people an idea about what goes where.
The only thing missing is some instruction on how to use it in README.

simi:
Yes, in README and CONTRIBUTING. Do you have any idea for copy for that?

scarfacedeb:
I think it should be something related to reduced test case.

> Isolate the demonstrable problem — Try to reproduce the problem with the Demo or with a reduced test case that includes the least amount of code necessary to reproduce the problem.


As an example: https://github.com/davatron5000/FitVids.js/blob/055bbd5a212e587ffba1604f08f75f572c3068b0/CONTRIBUTING.md

More general post about it: http://css-tricks.com/reduced-test-cases/

simi:
Rails text: http://guides.rubyonrails.org/contributing_to_ruby_on_rails.html#create-a-self-contained-gist-for-active-record-and-action-controller-issues

shioyama:
Hey sorry late to this, I'd be happy for any improvement to [CONTRIBUTING](https://github.com/globalize/globalize/blob/f1d7b06d53a5820a9e6421ae6f4357906d98292b/CONTRIBUTING.md), I just put it there a while back because we had no instructions at all at the time. The script seems like a reasonable place to start.

How about: "If you are filing a bug report, please consider adapting the provided template to reproduce the problem with minimal dependencies."

I like @scarfacedeb's text as well.

shioyama:
We should also update the existing contributing text to reflect changes to globalize versions and branches.

simi:
ping @shioyama

PR #417, opened 2015-02-17
Make presence validations work with fallback locales

### LINKED ISSUE #144: Trouble on updating an istance attribute when using ':fallbacks_for_empty_translations => true'
I use two locale languages: `en` and `de`. The default is `en`.

Given I have

```ruby
# Model file

translates :title, :fallbacks_for_empty_translations => true
validates :title, :presence  => true
```

When I run the following code

```ruby
# Controller file

I18n.locale
# => "en"
@article.inspect
# => #<Article id: 1, title: "Sample title">
params[:article].inspect
# => {"title"=>""}

@article.update_attributes(params[:article])
```

Globalize3 seems to work as expected and the `@article` is not updated (an error message is added to the `title` attribute related to the `presence` validation).

*However*, when I run the following code

```ruby
# Controller file

# I18n.locale
# => "de" # Note that the locale language is not the default 'en'.
# @article.inspect
# => #<Article id: 1, title: "Sample title">
# params[:article].inspect
# => {"title"=>""}

@article.update_attributes(params[:article])

# @article.inspect
# => #<Article id: 1, title: "Sample title">
```

Globalize3 seems to skip  the `validates :title, :presence  => true` for the `de` locale language and it updates the `@article` (even if no `title` value is provided) since it fallbacks to the `en` default translation (*no* error message is added to the `title` attribute related to the `presence` validation).

If I comment out the `:fallbacks_for_empty_translations => true` statement in the model file, it works as expected but I can not take any advantages from "fallbacking". So, *how can I add an error message to the `title` attribute related to the `presence` validation when I am not using the default locale language?*

*P.S.*: Maybe you can implement something that handles this common "situation"/"use case" relating to validation processes.

scarfacedeb:
I think I managed to resolve #144 issue by disabling fallbacks during save.

I added a test based on that issue.

@simi @parndt what do you think?

### DIFF
--- a/lib/globalize/active_record/instance_methods.rb
+++ b/lib/globalize/active_record/instance_methods.rb
@@ -149,7 +149,9 @@ module Globalize
 
       def save(*)
         Globalize.with_locale(translation.locale || I18n.default_locale) do
-          super
+          without_fallbacks do
+            super
+          end
         end
       end
 
@@ -193,6 +195,15 @@ module Globalize
           yield
         end
       end
+
+      def without_fallbacks
+        before = self.fallbacks_for_empty_translations
+        self.fallbacks_for_empty_translations = false
+        yield
+      ensure
+        self.fallbacks_for_empty_translations = before
+      end
+
     end
   end
 end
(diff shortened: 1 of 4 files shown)


parndt:
Seems good

PR #420, opened 2015-02-22
Add request_store to make Globalize thread safe with threaded web servers

### LINKED ISSUE #224: Globalize.locale remaining set over different sessions

scarfacedeb:
It adds `request_store` gem to resolve #224 issue and make Globalize thread-safe by default.
Code is basically a modification of @simi work in #413.

I added more threads to the multithreading test, because I noticed that test would randomly succeed with ivars, instead of `Thread.current` or `RequstStore.store` 1 out of 6 times.
Increasing number of threads to 5 resolved that issue and now I always get failure with ivars.

Unfortunately, I couldn't come up with a way to test #224 issue.
In theory, I could create a test that would reuse threads (in a similar way as puma and thin servers do it), but I think it would be too superfluous and fake.
The better test would be to run those servers and assert that `Globalize.locale` value is unset after every request, but it also seems too unnecessary, because it seems to me that we would be testing an external gem, instead of globalize code.

@simi @parndt what do you think?

### DIFF
--- a/globalize.gemspec
+++ b/globalize.gemspec
@@ -17,6 +17,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'activerecord', '>= 4.2.0', '< 4.3'
   s.add_dependency 'activemodel', '>= 4.2.0', '< 4.3'
+  s.add_dependency 'request_store', '~> 1.0'
 
   s.add_development_dependency 'database_cleaner', '~> 1.4.0'
   s.add_development_dependency 'minitest'
--- a/readme.md
+++ b/readme.md
@@ -327,6 +327,13 @@ def cache_key
 end
 ```
 
+## Thread-safety
+
+Globalize uses [request_store](https://github.com/steveklabnik/request_store) gem to clean up thread-global variable after every request.
+RequestStore includes a Railtie that will configure everything properly for Rails 3+ apps.
+
+If you're not using Rails, you may need to consult a RequestStore's [README](https://github.com/steveklabnik/request_store#no-rails-no-problem) to configure it.
+
 ## Tutorials and articles
 * [Go Global with Rails and I18n](http://www.sitepoint.com/go-global-rails-i18n/) - introductory article about i18n in Rails (Ilya Bodrov)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 5.1.0 (unreleased)
 
 * Replaced `after_` callbacks with `before_` callbacks and set `autosave: true` by default. [#341](https://github.com/globalize/globalize/pull/341) by [Andrew Volozhanin](https://github.com/scarfacedeb)
+* Add [RequestStore](https://github.com/steveklabnik/request_store) to make Globalize thread-safe again [#420](https://github.com/globalize/globalize/pull/420)
 
 ## 5.0.1 (2015-02-15)
 
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -1,10 +1,10 @@
+require 'request_store'
 require 'active_record'
 require 'patches/active_record/xml_attribute_serializer'
 require 'patches/active_record/query_method'
 require 'patches/active_record/serialization'
 require 'patches/active_record/uniqueness_validator'
 require 'patches/active_record/persistence'
-
 
 module Globalize
   autoload :ActiveRecord, 'globalize/active_record'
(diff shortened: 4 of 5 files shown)


globalize.gemspec
```diff
@@ -26,4 +27,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rdoc'
 
   s.add_development_dependency 'rake'
+  s.add_development_dependency 'pry-byebug'
```
simi: :scissors:

simi:
Looks good. I'm going to test this.

readme.md
```diff
@@ -327,6 +327,13 @@ def cache_key
 end
 ```
 
+## Thread-safety
+
+Globalize uses [request_store](https://github.com/steveklabnik/request_store) gem to clean up thread-global variable after every request.
+It includes a Railtie that will configure everything properly for Rails 3+ apps.
```
simi: s/It/RequestStore/g

readme.md
```diff
@@ -327,6 +327,13 @@ def cache_key
...
+
+If you're not using Rails, you may need to consult a request_store's README to configure it.
```
simi: link request store README here please

scarfacedeb:
@simi README is corrected, pry-rails was :scissors:

simi:
@scarfacedeb  I'm a little lost in request store's real name. But we need to unify it in README anyway.

* gemspec name uses [request_store](https://github.com/steveklabnik/request_store/blob/master/request_store.gemspec#L7)
* gemspec description and summary uses [RequestStore](https://github.com/steveklabnik/request_store/blob/master/request_store.gemspec#L11-L12)
* README.md uses [RequestStore](https://github.com/steveklabnik/request_store/blob/master/README.md#requeststore--)

I think RequestStore should be used. Can you confirm @steveklabnik?

steveklabnik:
It follows Ruby conventions, the name is RequestStore but that translates to request_store as a package name.

scarfacedeb:
@steveklabnik thanks!

@simi What do you think about this phrasing?

Globalize uses [request_store](https://github.com/steveklabnik/request_store) gem to clean up thread-global variable after every request.
RequestStore includes a Railtie that will configure everything properly for Rails 3+ apps.

If you're not using Rails, you may need to consult a RequestStore's [README](https://github.com/steveklabnik/request_store#no-rails-no-problem) to configure it.

simi:
looks good @scarfacedeb

scarfacedeb:
@simi it's commited.

simi:
Last thing, can you add CHANGELOG note please?

scarfacedeb:
Under which version should I add the changes?
5.1.0 (unreleased) or 5.2.0 (unreleased)?

simi:
5.1.0 please

globalize.gemspec
```diff
@@ -17,6 +17,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'activerecord', '>= 4.2.0', '< 4.3'
   s.add_dependency 'activemodel', '>= 4.2.0', '< 4.3'
+  s.add_dependency 'request_store', '~> 1.1'
```
simi: Isn't ~> 1.0 enough here?

scarfacedeb: Yes, you're right.
I'm used to defining it this way, because it gives me a hint about when I added that dependency. (it's more applicable in Gemfile). And to be honest, I've never even thought about specifying it your way.
I can change it, if you want.

simi: Just to make it as easy as possible to migrate from Globalize 5.0 to Globalize 5.1. There are only some minor changes from 1.0 to 1.1 in RequestStore codebase. But feel free to leave it as is now.

scarfacedeb: I see your point.
~> 1.0 means that less people would also have to update request_store from 1.0 to 1.1, if they already have it.
I changed it.

simi:
:confetti_ball:

scarfacedeb:
yay! :tada: 
@simi thank you!

mtparet:
Thank you for your work @scarefacedb @simi !

PR #421, opened 2015-02-22
Refactor translation_class to lookup constant only in the receiver

scarfacedeb:
The second parameter in `const_defined?` and `const_get` determines if it should lookup in the ancestors.
It prevents `const_get` from getting `::Translation` model, instead of `klass::Translation`.
I think that it makes this code a bit cleaner.

Related commit: https://github.com/globalize/globalize/commit/e0d1f57153ab97516665cb48f65f982c70d21433

### DIFF
--- a/lib/globalize/active_record/class_methods.rb
+++ b/lib/globalize/active_record/class_methods.rb
@@ -42,8 +42,9 @@ module Globalize
 
       def translation_class
         @translation_class ||= begin
-          klass = self.const_get(:Translation) rescue nil
-          if klass.nil? || klass.class_name != (self.class_name + "Translation")
+          if self.const_defined?(:Translation, false)
+            klass = self.const_get(:Translation, false)
+          else
             klass = self.const_set(:Translation, Class.new(Globalize::ActiveRecord::Translation))
           end
 


scarfacedeb:
Also, I wanted to completely remove `class_name` method, because it almost duplicates the output of  `model_name.singular`.
But I stumbled upon some issues.

`class_name` takes into account `self.table_name` attribute of the translated models, but `model_name` doesn't.
Moreover, with my changes, `class_name` is now used only in the `options[:foreign_key] ||= class_name.foreign_key` line inside `apply_globalize_options` method.
Given that, I think that it makes even more sense to get rid of explicitly defined `foreign_key` option on `has_many :translations` association and let rails derive it from the model name. (and it looks like that it also uses `model_name.singular`)

But it would some of the tests and introduce backward incompatibility.

An example from the tests:
```ruby
class NewsItem < ActiveRecord::Base
  self.table_name = :news
  translates :title, :foreign_key => :news_id
end
```

Now it results in `news_id` foreign_key for translations association.
But I remove the `foreign_key` option from the has_many call, it would use `news_item_id` foreign_key instead.
```ruby
[2] pry(main)> NewsItem.reflect_on_association(:translations).foreign_key
=> "news_item_id"
```

simi:
changelog please if something has changed

scarfacedeb:
@simi I hope not. Everything should have stayed the same.

PR #422, opened 2015-02-22
Derive foreign_key from Rails

scarfacedeb:
`class_name` method is used to derive `foreign_key` for associations from table names.

I think it's a good idea to remove it and allow rails itself to derive foreign_key value. (in cases when it's not defined by hand)

The problem with `class_name` is that its output diverges from the rails, because class_name uses table_name, instead of variation of model_name.

I removed `class_name`, but it broke tests that involve STI.
First of all, I added missing `type` columns to adhere to the latest rails conventions.

After that, I decided to refactor those models to illustrate 2 cases:
- When translates is called on parent STI model: `Media` and `Picture` models. 
- When translates is called on inherited STI model and foreign key is derived from that model: `Comment` and `TranslatedComment` models.

These changes required to rename foreign_key for `TranslatedComment` model from `comment_id` to `translated_comment_id`.

I think that it makes more sense, because we're only associating TranslatedComment records with translations,  not all of the comments. 
Also Rails is expecting this foreign key by default.

All of it unfortunately leads to backward incompatibility in cases when `transaltes` macro is called from inherited STI models.

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -13,13 +13,6 @@ module Globalize
         allow_translation_of_attributes(attr_names) if attr_names.present?
       end
 
-      def class_name
-        @class_name ||= begin
-          class_name = table_name[table_name_prefix.length..-(table_name_suffix.length + 1)].downcase.camelize
-          pluralize_table_names ? class_name.singularize : class_name
-        end
-      end
-
       def translates?
         included_modules.include?(InstanceMethods)
       end
@@ -42,7 +35,6 @@ module Globalize
 
       def apply_globalize_options(options)
         options[:table_name] ||= "#{table_name.singularize}_translations"
-        options[:foreign_key] ||= class_name.foreign_key
 
         class_attribute :translated_attribute_names, :translation_options, :fallbacks_for_empty_translations
         self.translated_attribute_names = []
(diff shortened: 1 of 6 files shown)


lib/globalize/active_record/act_macro.rb
```diff
@@ -42,7 +35,6 @@ def allow_translation_of_attributes(attr_names)
 
       def apply_globalize_options(options)
         options[:table_name] ||= "#{table_name.singularize}_translations"
-        options[:foreign_key] ||= class_name.foreign_key
```
parndt: Shouldn't this still be able to be specified in the case that we can't find it via reflection?

scarfacedeb: What do you by "we can't find it"?

We still can specify foreign_key by hand via `translates :title, foreign_key: 'page_id'`.
In case there's no `options[:foreign_key]` in `setup_translates!`, we call has_many first, to create association.

In any case, rails would create a reflection of `translations` association that would contain `foreign_key` - either set manually via `options` or derived magically by Rails.

And after that happened, it adds resulted foreign_key to translated_options anyway.

parndt: Oh, I see.. I just read the code in context and understand what you mean!

PR #430, opened 2015-03-25
Add UUID support

daniel-de-wit:
When the model that needs to be translated uses UUID as primary key, this pr will make the translation model also UUID. Setting the reference type is supported since 4.2 but if i'm correct the depency listen in the gemspec specifies 4.2

It's not perfect, but works.

Any ideas?

### DIFF
--- a/lib/globalize/active_record/migration.rb
+++ b/lib/globalize/active_record/migration.rb
@@ -67,8 +67,8 @@ module Globalize
         end
 
         def create_translation_table
-          connection.create_table(translations_table_name) do |t|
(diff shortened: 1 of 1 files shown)


franzliedke:
I assume this can be closed, as [a similar line seems to be in master](https://github.com/globalize/globalize/blob/6f9d3f38d132f3e5dcbe33311674e696c922d23c/lib/globalize/active_record/migration.rb#L81)?

franzliedke:
...but that hasn't been released, it seems. :(

PR #431, opened 2015-03-26
Unsaved records does not have any translations

runephilosof:
While a model is unsaved the translations method returns an empty array.
The attribute_translations returns the proper translations though.


parndt:
Thanks! Is this test just to prove the failure?

runephilosof:
Yeah. Instead of creating an issue, I thought it would be better show what fails this way.
Would it have been better as an issue?

parndt:
No, this is good, I was just checking. :)

PR #433, opened 2015-04-01
Rename readme.md to README.md

elia:
This is to let the gemspec pick up the file as it's way more useful to have in the .gem than e.g. `CONTRIBUTING.md`.


simi:
Actually readme.md is packed in gem, but I like this change to unify text file names.

simi:
thanks @elia!

elia:
Thanks!

I noticed this line in the gemspec of v4

      s.files        = Dir['{lib/**/*,[A-Z]*}']

Which only picks up uppercase files and seems to be still there in master

PR #437, opened 2015-04-23
To make uniqueness validator work with squeel (Rails 3.0)

torumori:
Please consider this to be fixed although 3-0-stable might be already out of support, but without this change, it causes ActiveRecord::StatementInvalid exception if the application uses squeel gem and the translated model uses the uniqueness validation.

https://github.com/activerecord-hackery/squeel#symbols-as-the-value-side-of-a-condition-in-normal-where-clauses
"If you have any where() clauses that use a symbol as the value side (right-hand side) of a condition, you will need to change the symbol into a string in order for it to continue to be treated as a value."

### DIFF
--- a/lib/patches/active_record/uniqueness_validator.rb
+++ b/lib/patches/active_record/uniqueness_validator.rb
@@ -31,7 +31,7 @@ ActiveRecord::Validations::UniquenessValidator.class_eval do
 
         sql, params  = mount_sql_and_params(finder_class, table_name, attribute, value)
 
-        relation = table.where(sql, *params).where(:locale => Globalize.locale)
+        relation = table.where(sql, *params).where(:locale => Globalize.locale.to_s)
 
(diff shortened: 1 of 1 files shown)


parndt:
I don't see a problem with this, except that the tests are failing but this doesn't seem to be your fault.

parndt:
@torumori which version are you using? Do you want a new gem release?

torumori:
@parndt Thank you for your quick response!
We are using version 3.0.4 globalize for Rails 3 version of our application. If you could provide a new release like 3.0.5 or 3.0.4.1, it would be greatly appreciated.

parndt:
@torumori thanks for fixing it!

Version 3.0.5 has been released to https://rubygems.org/gems/globalize/versions/3.0.5

torumori:
@parndt thank you much!

PR #441, opened 2015-05-11
This test is failing due to wrong processing columns with :null -> false constraint

xsrgx:


xsrgx:
This test is failing due to wrong processing columns with :null -> false constraint

simi:
Ahh, it was good enough to use gist for this, but thanks anyway. I'll try to dig deeper into this.

PR #446, opened 2015-06-16
Ensure order for translations

jdevega:
As some DB Engines do not ensure the order of the returned rows, it could happen that translations are not ordered correctly. So we have to ensure the correct sorting on the has_many association.

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -71,7 +71,8 @@ module Globalize
 
         translation_class.table_name = options[:table_name]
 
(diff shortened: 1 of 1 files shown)


parndt:
@jdevega just to confirm: this is still an issue after #447 ?
Looks like your change has resulted in a failing CI build. Can you take a look, please? :smile: 
```
ActiveRecord::StatementInvalid:         ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: post_translations.created_at: SELECT "post_translations".* FROM "post_translations" WHERE "post_translations"."post_id" = ?  ORDER BY "post_translations"."created_at" DESC
```

Issue #187, opened 2013-01-11
Add support for non-string types


dorner:
We are using globalize in two separate systems - one is the production system, where text is displayed; but we are also using it to enter that data, in an admin system. I added some magic functions to make this easier (e.g. for forms), but ran into a wall when I realized there is a shortcoming in the design.

We want to globalize not only string fields, but to allow us to store arbitrary fields per language. For example, there is a list of items (integer) in a flyer for each language. I'd like to be able to get/set that value per language using globalize3's functionality (i.e. number of English items, number of French items, etc.). We also have fields indicating when a piece of work is completed in each language (datetime). However, right now it seems that it only supports the :text and :string types. Is there anything in the design of globalize3 that would require this? What string-specific functionality is there?

I saw this pull request https://github.com/svenfuchs/globalize3/pull/116  but saw it was declined due to the use case not being valid. I hope this one is better described. 8-)

jamesneve:
+1
Is this still a likely feature to come out anytime in the near future? Like other people, I'm also having to use a rather awkward hack to make it work.

vizo:
:+1:

vizo:
I also need to save different longitude, latitude per language.

@jamesneve can you please show your hack?

slavajacobson:
Still nothing? :(

vizo:
It worked for me ....

You can do initializer globalize_override.rb and put inside:
```
module Globalize
  module ActiveRecord
    module Migration
      Migrator.class_eval do
        def valid_field_type?(name, type)
          # Allow translation of any type of field
          !translated_attribute_names.include?(name) || true # || [:string, :text].include?(type)
        end
      end
    end
  end
end

```

Dschee:
Thank you @vizo, I have just filed a pull requests that tries to fix this issue using your approach (by just allowing other types). My idea was to create a new feature branch named `non_charbased_type_support` but I couldn't find a way to create a new branch within a pull requests so the maintainers will need to do some manual work to merge that the intended way.

Once the feature branch is created, the feature could be tried out just by adding this to the Gemfile:
`gem 'globalize', github: "globalize/globalize', branch: 'feature/non_charbased_type_support'`

I hope this helps getting things started with this issue.

MartinDesigner:
![imghome](https://cloud.githubusercontent.com/assets/13809960/9289193/b888a334-4367-11e5-98a2-47387e3555cd.jpg)
[...]

PR #450, opened 2015-06-27
New branch to work on non char-based table column type support

Jeehut:
This request is only to get work started in order to support more data types than just `string` and `text`. The only thing I changed here is to remove the restriction to the two types and add some reference to this work within the README file. Referring to multiple users in the #187 issue the changes here should already work for most cases.

Of course this doesn't mean there's now full support for more types, but this should get us started and give people the option to try this branch out and see what the edge cases are so we can cover them. Here's some next steps to do:
- [ ] Write test code for each specific added type (integer, float, decimal, datetime, time, date, boolean)
- [ ] Figure out if `primary_key` and `binary` types should or could be enabled and enable them, if needed
- [ ] Update Documentation in README file to include numeric/date/boolean examples

I'm looking forward to your opinions. Thanks to @vizo for pointing towards the code to change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,12 @@ gem 'globalize', '~> 3.1.0'
 
 The [`3-1-stable` branch](https://github.com/globalize/globalize/tree/3-1-stable) of this repository corresponds to the latest ActiveRecord 3 version of globalize. Note that `globalize3` has been deprecated and you are encouraged to update your Gemfile accordingly.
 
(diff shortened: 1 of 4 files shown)


Dschee:
Please note that I couldn't find a way to create a new branch within a pull requests. So instead of merging this as is the maintainers may want to:

- [ ] Move the changes within the library to a new feature branch called `non_charbased_type_support`
- [ ] Merge the changed to the Readme file

Thank you for maintaining this great library!

PR #451, opened 2015-06-30
Inform about required translates call in the model before creating translations table

PragTob:
Without the translates call I got `undefined method `create_translation_table!' for #<Class:0x00000009124130>/` and when just specifying one of them I got `Missing translated field :banner_headline` - had to add them all for it to work. Information would be nice to have in the README so others don't make the same mistake as I did.

Thank you very much for your work :+1:


simi:
:heart:

PR #467, opened 2015-09-30
Add config option to globally enable fallbacks for empty translations

awernick:
I don't know if this is an edge case or not, but I added the option to enable fallbacks for empty translations for all models. It might not be necessary, but I thought it might be useful in some cases.

### DIFF
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -51,6 +51,15 @@ module Globalize
     def default_fallbacks(for_locale = self.locale)
       i18n_fallbacks? ? I18n.fallbacks[for_locale] : [for_locale.to_sym]
     end
+    
+    def fallbacks_for_empty_translations=(value = false)
+      return if !!value != value # Only allow booleans
+      storage[:globalize_fallback_empty_translations] = value
+    end
+
+    def fallbacks_for_empty_translations?
+      !!storage[:globalize_fallback_empty_translations]
+    end
 
     # Thread-safe global storage
     def storage
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -53,7 +53,7 @@ module Globalize
     end
     
     def fallbacks_for_empty_translations=(value = false)
-      return if !!value != value # Only allow booleans
+      return if [true, false].include? value
       storage[:globalize_fallback_empty_translations] = value
     end
 
--- a/lib/globalize/active_record/adapter.rb
+++ b/lib/globalize/active_record/adapter.rb
@@ -83,7 +83,7 @@ module Globalize
       end
 
       def fallbacks_for?(object)
-        object.nil? || (fallbacks_for_empty_translations? && object.blank?)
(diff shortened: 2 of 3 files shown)


phansch:
:+1: Ran into the same issue

coorasse:
:+1:

lib/globalize/active_record/adapter.rb
```diff
@@ -83,7 +83,7 @@ class << object; attr_accessor :translation_metadata end
       end
 
       def fallbacks_for?(object)
-        object.nil? || (fallbacks_for_empty_translations? && object.blank?)
+        object.nil? || (Globalize.fallbacks_for_empty_translations? && object.blank?) || (fallbacks_for_empty_translations? && object.blank?)
```
coorasse: `object.blank? && (Globalize.fallbacks_for_empty_translations? || fallbacks_for_empty_translations?)`

test/globalize/fallbacks_test.rb
```diff
@@ -93,6 +93,20 @@ class FallbacksTest < MiniTest::Spec
       assert_equal 'foo', post.title
     end
 
+    it 'resolves fallbacks with empty translations' do
+      Globalize.fallbacks_for_empty_translations = true
+      I18n.fallbacks.map :'de-DE' => [ :'en-US' ]
+      post = Post.create :title => 'foo'
```
coorasse: Use new hash syntax: `title: 'foo'`

test/globalize/fallbacks_test.rb
```diff
@@ -93,6 +93,20 @@ class FallbacksTest < MiniTest::Spec
...
+
+      I18n.locale = :'de-DE'
+      assert_equal 'foo', post.title
+
+      post.update_attributes :title => ""
```
coorasse: same here

lib/globalize.rb
```diff
@@ -51,6 +51,15 @@ def fallbacks(for_locale = self.locale)
     def default_fallbacks(for_locale = self.locale)
       i18n_fallbacks? ? I18n.fallbacks[for_locale] : [for_locale.to_sym]
     end
+    
+    def fallbacks_for_empty_translations=(value = false)
+      return if !!value != value # Only allow booleans
```
coorasse: I personally find `[true, false].include?  value` more readable and you can also get rid of the comment explaining what it does

parndt:
@coorasse good review!

test/globalize/fallbacks_test.rb
awernick: The only reason I used the old hashrocket syntax is because the other tests use it as well. Just trying to keep the code base consistent :+1:

awernick:
It's been almost a year, but I still believe this would be a good feature to have.

parndt:
@awernick thanks - would you please be able to write a note about it in the README (https://github.com/globalize/globalize/#i18n-fallbacks-for-empty-translations) so that others know about it too?

phansch:
I believe this is almost complete and still a desired feature?

parndt:
@phansch would you like to contribute the README for the config setting please? :smile: :heart:

PR #471, opened 2015-10-12
Clear schema cache before adding missing columns

faucct:
To make sure that the columns are really missing.


parndt:
Seems like a great idea, sorry this sat around for so long @faucct

PR #486, opened 2016-02-01
validate uniqueness of translated attributes

maatz:
It is possible to create duplicated translations, which should in my opinion not be possible.

E.g.
`5.times { |i| some_obj.translations.create!(locale: :en, title: "title ##{i}") }`
is possible, which will lead to 5 english translations for the same record.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@ Gemfile.lock
 *.gem
 .rbx
 .ruby-version
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Gemfile.lock
 *.gem
 .rbx
 .ruby-version
+.idea/
--- a/lib/globalize/active_record/translation.rb
+++ b/lib/globalize/active_record/translation.rb
@@ -4,6 +4,8 @@ module Globalize
 
       validates :locale, :presence => true
 
+      validate :uniqueness_of_translated_attribute, if: proc { combination_changed? }
+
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ Gemfile.lock
 *.gem
 .rbx
 .ruby-version
+.idea/
```
simi: https://help.github.com/articles/ignoring-files/#create-a-local-gitignore

tomash:
Looks good to me (except for the `.gitignore` entry) :+1:

ridiculous:
I'm against having this as a default. What about adding this as a configuration option or plugin?

parndt:
I think this would lead to better data - what's your main objection around it @ridiculous ?

ridiculous:
My main objection is that it will do a read before every insert. Furthermore, this isn't needed by everyone who uses Globalize. For instance, uniqueness could be ensured farther upstream or it may not be a concern in their system. I ran into this uniqueness issue on an app I work on, which is why I opened https://github.com/globalize/globalize/pull/521. I'd prefer the uniqueness constraint be implemented using a database index. 

That being said, I'm still open to this implementation, I just don't think it should be the default behavior.

dalpo:
Personally I approve the idea of @ridiculous

parndt:
I agree, unique constraints should be achieved with a database index.  Regardless, thanks for the proposal!

PR #438, opened 2015-04-27
Add migration locale option

foxtacles:
This adds a new optional parameter to the options hash in `create_translation_table!` to specify the target locale for existing data.

I personally found this useful for our projects. Feel free to comment :)

PR #494, opened 2016-03-09
Use I18n.locale instead of default_locale when migrating data

ybart:
It would allow to choose the locale used in migration like this:

```ruby
I18n.with_locale(:bo) do
  Product.create_translation_table!({ name: :string, values: { type: :string, array: true }, hint: :text }, migrate_data: true)
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,19 @@ end
 
 NOTE: Make sure you drop the translated columns from the parent table after all your data is safely migrated.
 
+In order to use a specific locale for migrated data, you can use `I18n.with_locale`:
+
+```ruby
+    I18n.with_locale(:bo) do
+      Post.create_translation_table!({
+        :title => :string,
+        :text => :text
+      }, {
+        :migrate_data => true
+      })
+    end
+```
+
 ## Adding additional fields to the translation table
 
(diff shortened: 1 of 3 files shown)


ybart:
See also : https://github.com/globalize/globalize/pull/438

parndt:
@ybart would you prefer that #438 was merged instead of this PR?

ybart:
I tend to prefer this PR because it uses standard I18n syntax. Furthermore it's only a single change in the code. However, it might need tests and documentation. If you'd rather merge this one instead of #438, I can add tests and documentation for this if needed.

However if you prefer #438 syntax I see no problem closing this one as I just opened it as a proposal.

parndt:
Thanks @ybart 

@foxtacles sorry, I know it's been a while, but which one do you prefer?

foxtacles:
Hey, I think this one is clearly better. Not sure why I haven't thought of it myself :)

I'm going to close my PR once this one has documentation/tests/whatever else is needed.

ybart:
I'll take care of it.

ybart:
It should be ready to merge.

parndt:
Excellent!

PR #496, opened 2016-03-10
Add support default values 

ybart:
When default value is specified for a given column in a migration, when there is no pre-existing translation on a record, trying to get the attribute will return nil instead of the default value.

The migration:

```ruby
Product.create_translation_table!(values: { type: :string, array: true, null: false, default: [] })
```

Currently, we get:

```ruby
Product.new.values
=> nil
```

I would rather expect:

```ruby
Product.new.values
=> []
```

With this PR, I get:

```ruby
Product.new.values
=> "{}"
```

This is better but needs a conversion, I did not figure how to do it yet.

### DIFF
--- a/lib/globalize/active_record/adapter.rb
+++ b/lib/globalize/active_record/adapter.rb
@@ -68,7 +68,13 @@ module Globalize
 
       def fetch_attribute(locale, name)
         translation = record.translation_for(locale, false)
-        return translation && translation.send(name)
+        return translation ? translation.send(name) : default_value(name)
+      end
+
+      def default_value(name)
+        column = column_for_attribute(name)
+
+        column.type_cast_from_database(column.default)
       end
 
       def set_metadata(object, metadata)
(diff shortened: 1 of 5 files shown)


parndt:
Hello @ybart, thanks for your contribution!  It would be great if you could write some supporting tests for this new feature please; is that a possibility?

ybart:
This should not be a problem. I'm doing that today.

ybart:
Code updated, rebased, and tests added.

I've improved the code so it behaves as expected with PG and added tests (returning `[]` rather than `"{}"`). The improvement shows no difference with SQLite3 (which returns `"--- []\n"` in all cases).

Since the tests are run only against SQLite3, I only tested the SQLite3 behavior.

By the way, I noticed trying to use it that `#type_cast` vas not defined for `ActiveRecord::ConnectionAdapters::Column` in AR 4.2 so I believe that `Globalize::ActiveRecord::Adapter#type_cast` is likely to be dead code.

parndt:
Our continuous integration works for PG and MySQL as well :smile:

ybart:
I did not realize that. Thanks !

PR #499, opened 2016-04-15
Track globalize_serialized_attributes per class and not globally

### LINKED ISSUE #472: Bug in Rails 4 when using translate in combination with serialize
In our app we have two models `A` and `B`, both having a column `data` which is a translated attribute. In addition to that, `B` also serializes the `data` column: 

``` ruby
class A < ActiveRecord::Base
  translates :data
end

class B < ActiveRecord::Base
  serialize :data, Hash
  translates :data
end
```

We run into some really strange behaviour: The configuration of `B` seems to affect `A`, because when we try to save an `A` instance to the database, we get the following error: 

`ActiveRecord::SerializationTypeMismatch (Attribute was supposed to be a Hash, but was a String. ... )` 

`A` isn't supposed to be serializing anything, so it looks like the configuration on `B` is messing up `A`.

The problem goes away if we remove the `translate` configurations on both Models.

ebeigarts:
mattr_accessor doesn't work well with inheritance so class_attribute is used instead.
class_attribute doesn't work well with mutable objects so a new hash is set when hash gets modified.

Fixes #472.

### DIFF
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -84,7 +84,7 @@ module Globalize
   end
 end
 
-ActiveRecord::Base.mattr_accessor :globalize_serialized_attributes, instance_writer: false
+ActiveRecord::Base.class_attribute :globalize_serialized_attributes, instance_writer: false
 ActiveRecord::Base.globalize_serialized_attributes = {}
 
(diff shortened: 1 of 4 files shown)


parndt:
This passes CI: https://travis-ci.org/globalize/globalize/builds/263090063

Thanks!

PR #501, opened 2016-04-22
Patch query_method until Rails4

bricesanchez:
This PR should fix the `number?` error : https://github.com/globalize/globalize/pull/485#issuecomment-200804762

PR #503, opened 2016-04-26
Type cast fix rails5

jdbwin:


jdbwin:
Will send another pr when #501 is closed

PR #504, opened 2016-04-27
Enhance association to translations

shelling:
- Use rails convention on translation_class.belongs_to.
  `Post::Translation` will belong to `:post`.
  `picture.translations.first.picture` will be a `Picture`.
- Use `inverse_of` on both sides to improve performance.
  This is also helpful when trying to validate presence of
  association when both sides are not persisted yet.
- `#globalized_association` can calculate association name.
  `Post::Translation` will use `:post` to belong to.
  Fall back to `:globalized_model` if association model is reserved.
  `Locale::Translation` will use `:globalized_model` to belong to.
- Fix bugs in `#set_translations`.
  `post = Post.new(title: "Title")` will save title in `globalize.stash`
  `post.set_translations(title: "Updated Title")` will create a translation
  but ignore values saved in `globalize.stash` and then invoke
  `globalize.save_translations!` which use values in `globalize.stash`
  to create another translation. The translation created by
  `set_translations(title: "Updated Title")` will not be saved.
  New implementation uses values from `globalize.stash` to create the
  translation and remove values from `globalize.stash`. Only one
  translation will be created and saved

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -72,6 +72,7 @@ module Globalize
 
         has_many :translations, :class_name  => translation_class.name,
                                 :foreign_key => options[:foreign_key],
+                                :inverse_of  => translation_class.globalized_association,
                                 :dependent   => :destroy,
                                 :extend      => HasManyExtensions,
                                 :autosave    => true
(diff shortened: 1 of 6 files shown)


shioyama:
Thanks for these changes. I'd like to merge them but seems rails5 tests are failing. We'll try to have a look at it so we can merge but if you could fix that, we'd get this merged faster.

PR #485, opened 2016-02-01
Rails 5 compatibility

pranik:
Moving globalize to rails 5.

This is my first pull requests so forgive me if i miss something.

# Incompatibility

I found some incompatibility between rails 4 and rails 5

- [X] activemodel-serializers-xml extracted from rails
- [X] where_values_hash moved to ActiveRecord::Relation
- [X] deprecated alias_method_chain with Module.prepend
- [X] use Module.prepend on Globalize patches
- [X] uniqueness_validator: build_relation return ActiveRecord::Relation so we cannot use arel methods
- [X] deprecation on passing parameters on destroy_all
- [X] rails 4 compatibility
- [x] update README
- [x] fix all tests

# Rails 4 compatibility

I check `ActiveRecord::VERSION:STRING` to load right module.

# Fix all tests

I have 2 test not working 

```
Minitest::Assertion: Expected: 1
  Actual: 2
  test/globalize/attributes_test.rb:31:in block (2 levels) in <class:AttributesTest>'
```

```
Minitest::UnexpectedError: ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: scoped_validatees.integer: SELECT  1 AS one FROM "scoped_validatee_translations" WHERE "scoped_validatee_translations"."string" = 'c' AND "scoped_validatee_translations"."locale" = ? AND "scoped_validatees"."integer" = ? AND "scoped_validatee_translations"."scope_string" = ? LIMIT ?

lib/patches/active_record/rails5/uniqueness_validator.rb:35:in `validate_each'
test/globalize/validations_test.rb:129:in `block (2 levels) in <class:ValidationsTest>'
```

I think this PR can be a good start point to add rails 5 compatibility.

PR #505, opened 2016-04-28
Fixing Issues 501, 502, 503

pranik:
Commits merged just before #485 broke travis build.

- [X] Fixed quoting error on Array class (rails 4 didn't raise TypeError).
- [X] Removed deprecated alias_method_chain introduced on DirtyAdapter
- [ ] Fixing remaing errors on dirty adapter only on rails 5

### DIFF
--- a/lib/globalize/active_record/adapter.rb
+++ b/lib/globalize/active_record/adapter.rb
@@ -68,13 +68,22 @@ module Globalize
 
       def fetch_attribute(locale, name)
         translation = record.translation_for(locale, false)
-        return translation ? translation.send(name) : default_value(name)
-      end
-
-      def default_value(name)
-        column = column_for_attribute(name)
-
-        column.type_cast_from_database(column.default)
+        if translation then
+          translation.send(name)
+        else
+          column = column_for_attribute(name)
+          if column.respond_to?(:type_cast_from_database)
+            column.type_cast_from_database(column.default)
+          else
+            type = record.class.translation_class.type_for_attribute(name)
+            default = type.deserialize(column.default)
+            if default.nil?
+              "-> { #{column.default_function.inspect} }" if column.default_function
+            else
+              default
+            end
+          end
+        end
       end
 
       def set_metadata(object, metadata)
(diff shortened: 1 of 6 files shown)


lib/globalize/active_record/adapter.rb
```diff
@@ -68,13 +68,22 @@ def unserializable_attribute?(name, column)
...
       def fetch_attribute(locale, name)
         translation = record.translation_for(locale, false)
-        return translation ? translation.send(name) : default_value(name)
-      end
-
-      def default_value(name)
-        column = column_for_attribute(name)
-
-        column.type_cast_from_database(column.default)
+        if translation then
```
parndt: We probably don't need `then`

parndt:
@pranik looks like you'll have to move your patch on query_methods to the new file https://github.com/globalize/globalize/blob/master/lib/patches/active_record/rails4/query_method.rb since https://github.com/globalize/globalize/blob/master/lib/patches/active_record/query_method.rb is different

lib/patches/active_record/quoting.rb
```diff
@@ -0,0 +1,21 @@
...
+              "'#{quote_string(YAML.dump(value))}'"
+            else
+              super
+          end
+        end
+      end
+    end
+  end
+
+  ActiveRecord::ConnectionAdapters::AbstractAdapter.include Globalize::ConnectionAdapters::Quoting
```
dalpo: I think that the issue about the `'_quote': can't  quote Array (TypeError)` is related to the test suite based only to the sqlite which doesn't support array datatypes. 
See: https://github.com/globalize/globalize/blob/master/test/data/schema.rb#L52

So I'm no sure that monkey patching the adapter is the right solution in this case.. But I need more time to investigate. :)

pranik: I think there is a problem on travis configuration. It seems that does'nt change the database adapter during testing and uses always sqlite. 
I expect no mistake when it uses postgres or mysql.

pranik:
Ok, I close this PR and I'll open a new one rebased with current master to avoid conflicts. Ok?

lib/patches/active_record/quoting.rb
parndt: That's right, I removed the DB constants from the environment because it turns out we were just forcing sqlite3 anyway so it was disingenuous.

parndt:
@pranik sure, I added you as a collaborator so feel free to just make a branch off the globalize master branch on this repository 😄

PR #506, opened 2016-05-03
Update migration instructions for newer Rails

abuisman:
The rails guides state it is preferable to use `reverisble`: http://edgeguides.rubyonrails.org/active_record_migrations.html#using-reversible


parndt:
Nice! @abuisman which version of Rails was this added?

parndt:
Cool - this has been around for a while https://github.com/rails/rails/commit/99770e4c659013461fb308040e3d594f2038ed24

PR #512, opened 2016-06-05
Rails 5 support work

parndt:
Thanks to @pranik 

I've opened this so that everyone can see the progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ sudo: false
 script: "bundle exec rake"
 gemfile:
   - Gemfile
-  - gemfiles/Gemfile.rails5
+  - gemfiles/Gemfile.rails4
+  - gemfiles/Gemfile.activemodel-serializers-xml
 rvm:
   - 2.3.1
   - 2.2.5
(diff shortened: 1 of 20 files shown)


pranik:
Now we have this error `ActiveModel::MissingAttributeError:ActiveModel::MissingAttributeError: can't write unknown attribute 'title'`due to `restore_attribute! `.

pranik:
Now we have this error on associations.
`test_0001_create on associations works#translated record::associations
ActiveModel::MissingAttributeError: missing attribute: content`

To load target related to association, rails call` merge_target_lists` that use `attribute_names` to detect record attributes.

```ruby
# active_record/associations/collection_association.rb
def merge_target_lists(persisted, memory)
  return persisted if memory.empty?
  return memory    if persisted.empty?

  persisted.map! do |record|
    if mem_record = memory.delete(record)

      ((record.attribute_names & mem_record.attribute_names) - mem_record.changes.keys).each do |name|
        mem_record[name] = record[name] # <<< this call: read_attribute(attr_name) { |n| missing_attribute(n, caller) }
      end

      mem_record
    else
      record
    end
  end

  persisted + memory
end
```

Globalize override attribute_names to return also translated attributes.
```ruby
# globalize/active_record/instance_methods.rb
def attribute_names
  translated_attribute_names.map(&:to_s) + super
end
```

Then rails try to `read_attribute` passing a block. During test `content` is nil and so call we obtain a  `missing_attribute`

```ruby
# active_record/attribute_methods
def [](attr_name)
  read_attribute(attr_name) { |n| missing_attribute(n, caller) }
end
```

I can fix removing `attribute_names` override but I don't know if there is a better solution.

pranik:
Any ideas how to fix `ArgumentError:         ArgumentError: method 'select_all': given 4, expected 4` in *rbx* tests?

parndt:
I'm not too worried if it is only rbx failing

parndt:
Turns out the rbx on Travis is just outdated, and it passes using the more recent version locally.

parndt:
Let's just get it into master!

PR #517, opened 2016-06-14
fix order by hash with string keys

davekaro:
useful when passing order params from query params like

```
Person.order(params[:order] => params[:direction])
```

### DIFF
--- a/lib/globalize/active_record/query_methods.rb
+++ b/lib/globalize/active_record/query_methods.rb
@@ -94,7 +94,7 @@ module Globalize
       def parse_translated_order(opts)
         case opts
         when Hash
-          ordering = opts.map do |column, direction|
+          ordering = opts.symbolize_keys.map do |column, direction|
             klass = translated_column?(column) ? translation_class : self
             klass.arel_table[column].send(direction)
           end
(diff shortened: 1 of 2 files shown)


parndt:
Couldn't you just pass them like this:

```ruby
Person.order(params[:order].to_sym => params[:direction])
```

This avoids forcing the library to symbolize all keys?

davekaro:
That is exactly what I did... Just found it surprising it didn't work out of the box. It seems like a common case to order using query params which will always be strings. Maybe just adding something to the README would be more appropriate?

parndt:
I'm of the opposite viewpoint, actually.  I don't think it's a great idea to be passing user input straight into database queries.  I'd run it through some sort of sanitisation first which would handle choosing the right symbol for me.

davekaro:
For more background, I was using the administrate gem and found that ordering didn't work because it was a translated field. See https://github.com/thoughtbot/administrate/blob/master/app/controllers/administrate/application_controller.rb#L87 and https://github.com/thoughtbot/administrate/blob/master/lib/administrate/order.rb#L10 for reference.

Furthermore, in Rails it doesn't matter if I pass a hash to `#order` with strings as keys or symbols as keys - it works the same both ways. Shouldn't I expect ordering to work the same when using translations too?

parndt:
Thanks for the links @davekaro 

@globalize/collaborators this isn't something I want to add to the core, but does anyone want to override that opinion?  I just don't think we should be symbolizing arbitrary user input.

davekaro:
Not that I know anything about the globalize internals... but alternatively maybe we could just convert https://github.com/globalize/globalize/blob/master/lib/globalize/active_record/query_methods.rb#L110 to something like:

```
translated_attribute_names.flat_map { |name| [name, name.to_s] }.include?(column)
```

PR #521, opened 2016-06-24
Add :unique_index option to #create_translation_table!

ridiculous:
What do you think about supporting a new `:unique_index` option that can be passed to the `Model.create_translation_table!` method to create a new unique index on `locale` and `model_id`?

### DIFF
--- a/lib/globalize/active_record/migration.rb
+++ b/lib/globalize/active_record/migration.rb
@@ -29,6 +29,10 @@ module Globalize
         end
 
         def create_translation_table!(fields = {}, options = {})
+          extra = options.keys - [:migrate_data, :remove_source_columns, :unique_index]
+          if extra.any?
+            raise ArgumentError, "Unknown migration #{'option'.pluralize(extra.size)}: #{extra}"
+          end
           @fields = fields
           # If we have fields we only want to create the translation table with those fields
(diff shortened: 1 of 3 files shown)


ridiculous:
Looks like Travis is failing on an unrelated Rails 5 thing..

parndt:
I think it would be best if this is handled using a manually created database migration, so that we can avoid the extra complexity.  Is that okay, @ridiculous ?

ridiculous:
Yeah, that's what we've been doing. It's a hassle though, cause we gotta drop the index added by Globalize and then re-add the same one, but with the unique constraint. Is really that more complex?

ridiculous:
Even if we don't support the new `:unique_index` option I'm proposing, I think we should still validate the given options, for the purpose of being more user friendly

parndt:
@ridiculous ah I didn't understand you had to drop and re-create it.. gotcha. Then this might be worthwhile after all!

lib/globalize/active_record/migration.rb
```diff
@@ -178,6 +191,14 @@ def clear_schema_cache!
 
         private
 
+        def truncate_index_name(index_name)
```
parndt: 👍

PR #524, opened 2016-07-01
Do not load Globalize::XmlSerializer if ActiveRecord::XmlSerializer is not present

rhymes:
Rails 5 has moved XML serializers in their own separate gem here

https://github.com/rails/activemodel-serializers-xml

globalize should work also if the Rails app does not use that gem


parndt:
@rhymes I've merged your work to master as part of #512 - thank you!

rhymes:
@parndt thanks!

PR #532, opened 2016-07-19
FIXED: 'false' values are returned as nil

florianeck:
when a globalized attribute was 'false', the return value was nil.

See Code for details.


parndt:
Thanks 😄

PR #535, opened 2016-07-28
Improve migration

tvcam:
- Allow reference to integer/string type
- Fix data migration on drop_table with dynamic model primary_key name

  e.g. Model with uuid:string/id:integer as primary key


tvcam:
Seems like @parndt is so busy right now.

parndt:
Sorry - makes sense not to hardcode `id` - thanks!

PR #537, opened 2016-08-04
Fix destroying attributes in with_given_locale

nobuhikosawai:
### Summary

This fixes the behaviour of  arguments of ```assign_attributes```.
The original code destroy the arguments after assign_attributes.
If you use the same parameter after assign_attriubtes, it causes a lot of trouble. 

Given a case you have a User model applied globalize, the following problem happens.
```rb
params = {"id" => 1, "name" => "John Doe"}
user = User.find_or_initilized_by(id: params["id"])
user.assign_attributes(params)
params["id"]  # returns nil, whereas params[:id] = 1
```

I use this code in my project, and it made system problems.
This problem derived from the change in arguments in the with_given_locale method. 
This PR remains the original parameters and make a copy instead.

(This is my first pull request so allow me if something is wrong.)

### DIFF
--- a/lib/globalize/active_record/instance_methods.rb
+++ b/lib/globalize/active_record/instance_methods.rb
@@ -12,11 +12,11 @@ module Globalize
       end
 
       def attributes=(attributes, *args)
-        with_given_locale(attributes) { super }
+        with_given_locale(attributes) { super(attributes_without_locale(attributes)) }
       end
 
       def assign_attributes(attributes, *args)
-        with_given_locale(attributes) { super }
+        with_given_locale(attributes) { super(attributes_without_locale(attributes)) }
       end
 
 
@@ -172,6 +172,10 @@ module Globalize
 
     protected
 
+      def attributes_without_locale(attributes)
+        attributes.try(:except, :locale)
+      end
+
       def each_locale_and_translated_attribute
         used_locales.each do |locale|
           translated_attribute_names.each do |name|
@@ -192,7 +196,7 @@ module Globalize
       end
 
       def with_given_locale(attributes, &block)
-        symbolized_attributes = attributes.symbolize_keys if attributes.respond_to?(:symbolize_keys!)
+        symbolized_attributes = attributes.symbolize_keys if attributes.respond_to?(:symbolize_keys)
 
         if locale = symbolized_attributes.try(:delete, :locale)
           Globalize.with_locale(locale, &block)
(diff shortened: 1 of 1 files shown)


parndt:
Yes, I agree that we shouldn't be mutating state in something like params.

Well done on your first pull request!!! 🎉

lib/globalize/active_record/instance_methods.rb
```diff
@@ -192,9 +196,9 @@ def save_translations!
       end
 
       def with_given_locale(attributes, &block)
-        attributes.symbolize_keys! if attributes.respond_to?(:symbolize_keys!)
+        symbolized_attributes = attributes.symbolize_keys if attributes.respond_to?(:symbolize_keys)
```
parndt: I have a small optimisation which ensures that the variable is always set, even if attributes doesn't respond to the method:

```ruby
def with_given_locale(_attrs, &block)
  attributes = _attrs.dup
  attributes.symbolize_keys! if attributes.respond_to?(:symbolize_keys!)

  if locale = attributes.try(:delete, :locale)
    Globalize.with_locale(locale, &block)
  else
    yield
  end
end
```

I stubbed out part of it and tried it in irb and it still works:

```
[29] pry(main)> a
=> {"locale"=>"foo"}
[30] pry(main)> def with_given_locale(_attrs, &block)
[30] pry(main)*   attributes = _attrs.dup
[30] pry(main)*   attributes.symbolize_keys! if attributes.respond_to?(:symbolize_keys!)
[30] pry(main)*   p attributes
[30] pry(main)*   if locale = attributes.try(:delete, :locale)
[30] pry(main)*     "Globalize.with_locale(locale, &block)"
[30] pry(main)*   else
[30] pry(main)*     "yield"
[30] pry(main)*   end
[30] pry(main)* end
=> :with_given_locale
[31] pry(main)> with_given_locale(a)
{:locale=>"foo"}
=> "Globalize.with_locale(locale, &block)"
[32] pry(main)> a
=> {"locale"=>"foo"}
```

Note how it symbolized the keys on the duplicated object inside the method scope but outside it the object still has the same string keys.

@nobuhikosawai What do you think?

nobuhikosawai: @parndt 
Thank you very much for your comment! I totally agree with your opinion.
I followed your suggestion and optimised my code!

parndt:
Wonderful! Thank you very much 👍🏼

nobuhikosawai:
I'm glad that I could contribute! 👍

PR #542, opened 2016-08-22
Simple English

PascalPixel:
I like to think I’m not an idiot, alas, sometimes in vain.
It took me reading half the linked-to tutorial to actually understand what this gem does.

I added a simple english version to the readme that even I would have understood.


parndt:
👍

PR #549, opened 2016-09-15
Add 'optional: true' to belongs_to association

crisan-ciprian:
This fixes rails 5 belongs_to being required by default:

```
class Course < ApplicationRecord
  translates :title
end

course = Course.new(title: 'title')
course.save  # returns false
course.errors.full_messages  # returns Translations globalized model must exist
```


crisan-ciprian:
@simi I think you are right. But for some reason I still get the problem mentioned in the description. I will close this pull request as I just noticed that there is another one trying to fix the same problem.

PR #546, opened 2016-09-11
add inverse_of to translations association

jdugarte:
This fixes a validation error:

```
class Position < ApplicationRecord
  translates :name
end

> p = Position.new
> p.name = 'Position 1'
> p.valid?
false
> p.errors.full_messages
[
    [0] "Translations globalized model must exist"
]
```

PR #551, opened 2016-09-24
Add inverse_of to translations association (with passing test)

shioyama:
Continuing from #546. [This](https://github.com/globalize/globalize/blob/a7a33088e86b28e83fb296999672cfa7ba24e090/test/globalize/set_translations_test.rb#L33) spec of `set_translations` was failing because it sets an attribute on the translation, but the globalized model takes precedence. Haven't looked into it in any depth but this seems to fix it:

```ruby
options[locale].each do |key, value|
  translation.send :"#{key}=", value
  translation.globalized_model.send :"#{key}=", value
end
```

I just set the same attribute on `translation.globalized_model`, which passes the test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,8 @@ matrix:
       gemfile: Gemfile
     - rvm: rbx
       gemfile: gemfiles/Gemfile.activemodel-serializers-xml.rb
+    - rvm: rbx
+      gemfile: gemfiles/Gemfile.rails-4.2.rb
     - rvm: jruby-9.1.2.0
       gemfile: Gemfile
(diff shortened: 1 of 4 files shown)


shioyama:
Just one build is failing on something unrelated to this PR.

shioyama:
@globalize/core I moved rbx entirely to allowed failures, following discussions with @brixen. We should aim to move to Trusty asap, but this PR is entirely unrelated and essential for Rails 5 support so I'd like to merge it first, then deal with build issues separately.

simi:
:tada: thanks!

PR #552, opened 2016-09-25
Switch to Trusty on travis

shioyama:
Builds are failing on postgres/rbx combination only. Proposed solution is to use latest version of rubinius:

Ref: https://twitter.com/shioyama/status/779813304713646081

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,12 @@ cache: bundler
 sudo: required
 dist: trusty
 before_install: gem install bundler
+addons:
+  apt:
+    packages:
+    - mysql-server-5.6
+    - mysql-client-core-5.6
+    - mysql-client-5.6
 before_script: bundle exec rake db:create db:migrate
 script: bundle exec rake
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ dist: trusty
 before_install:
   - gem install bundler
   - mysql -u root -e 'create database globalize_test character set utf8;' >/dev/null
+install:
+  - bundle install --retry=3
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,9 @@ language: ruby
 cache: bundler
 sudo: required
 dist: trusty
-before_install: gem install bundler
+before_install:
+  - gem install bundler
+  - mysql -u root -e 'create database globalize_test character set utf8;' >/dev/null
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 cache: bundler
-sudo: false
+sudo: required
+dist: trusty
 before_install: gem install bundler
 before_script: bundle exec rake db:create db:migrate
 script: bundle exec rake
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,12 @@ rvm:
   - jruby-9.1.2.0
 
 env:
+  matrix:
   - DB=sqlite3
   - DB=postgres
   - DB=mysql
+  global:
+    - "JRUBY_OPTS='--client -J-XX:+TieredCompilation -J-XX:TieredStopAtLevel=1 -Xcext.enabled=false -J-Xss2m -Xcompile.invokedynamic=false'"
 
 matrix:
   allow_failures:
--- a/test/support/database.yml
+++ b/test/support/database.yml
@@ -4,6 +4,7 @@ mysql:
   username: root
   hostname: localhost
   encoding: utf8
+  collation: utf8_general_ci
 
 postgres:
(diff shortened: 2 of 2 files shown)


shioyama:
This is a positively painful experience. If anybody with any idea how to get rubinius builds to work on Travis CI, please let me know.

shioyama:
@brixen Trying to get Rubinius to work again, updated to use Trusty and fixed issues with other builds but rbx 3.60 is still failing... will keep at it, let me know if you can provide any hints :smile:

shioyama:
Correction, the mysql build seems to be working, but sqlite and postgres are failing on `bundle install`...

.travis.yml
```diff
@@ -2,7 +2,9 @@ language: ruby
 cache: bundler
 sudo: required
 dist: trusty
-before_install: gem install bundler
+before_install:
+  - gem install bundler
+  - mysql -u root -e 'create database globalize_test character set utf8;' >/dev/null
```
dalpo: What do think about to put the  `character set utf8` argument directly here? 
https://github.com/globalize/globalize/blob/master/test/support/database.rb#L58

shioyama: Sure makes sense. I just want to figure out why rubinius builds are failing...

shioyama:
Ok I give up. I can't get rubinius to work and with the update to Trusty jruby doesn't seem to `bundle` either...

simi:
@shioyama don't worry, I try to take a look also.

brixen:
@shioyama I opened https://github.com/globalize/globalize/pull/554 to work on Rubinius compatibility, not to replace this PR but so I can separate updates from other changes you may need to make for other engines (eg why does JRuby not bundle on Trusty? Is that a blocker to using Trusty?)

For this PR, could you revert the change to using `rbx-3.60`? Just using `rbx` should work and using a more specific value will inhibit using newer releases as I fix issues.

shioyama:
@brixen Thanks for all your investigating. I removed the commit switching to rbx-3.60.

About JRuby, yes I believe that would be a blocker for switching to Trusty.

I'll look into #554 later today.

brixen:
@shioyama quick update, I was able to get building Rubinius Precise binaries working again https://travis-ci.org/rubinius/rubinius-build/builds/162983586

shioyama:
@brixen Brilliant, thanks!

brixen:
@shioyama omg, sorry for the tease. Unfortunately, using the LLVM-built packages requires using an alternate build of GCC and libc++, which isn't installed by default on Precise, so the binary tries to load a shared lib version of libatomic, which fails. :( I'm trying to figure out if I can statically link.

brixen:
@shioyama ok, bad news. I've now spent a few hours trying to get binaries to build for Precise and statically link the necessary libraries. I have not been successful and I'm not going to spend any more time. I'd suggest moving all Rubinius to `allow_failures` for Precise.

It's a tragedy of our collective infrastructure that Precise is still in use at all. If I understand Ubuntu LTS, it's only got a little over 6 months remaining. Orgs like Heroku have long since moved forward. Whatever the issue is with JRuby and Bundler on Trusty, it will need to be resolved sooner than later completely independent of anything Rubinius related.

I'll leave #554 open and continue to work on getting Rubinius green on Trusty. If this doesn't work for you, I understand.

Packaging and distribution has been an utter nightmare for Rubinius given the intricate, obscure, and painful interdependencies that exist in our OS infrastructure, like we see here. I've recently removed support for 32-bit, GCC, and won't support platforms for which no LLVM packages exist. This obviously limits the platforms that we support, but the alternative (supporting all kinds of legacy systems with multi-year out-of-date packages) becomes more untenable by the day.

Thanks for understanding and sorry for the frustrating experience.

brixen:
@shioyama and RVM is again failing to find the Rubinius 14.04 binaries despite that they are where they they have always been (see links in https://twitter.com/brixen/status/780979102857572352).

shioyama:
@brixen No apologies needed! Thank *you* for taking so much time to address this issue, to the point of actually looking at the gem dependencies. I think you're absolutely right that we'll have to deal with the transition to Trusty eventually so now's as good a time as any. I can take it from here, thanks again!

parndt:
@shioyama should we just move this to allowed failures?

shioyama:
@parndt rbx is already in [allowed failures](https://github.com/globalize/globalize/blob/master/.travis.yml#L28-L33). You mean switch to trusty with rbx on allowed failures? I guess since we're already allowing them to fail it wouldn't change much.

parndt:
@shioyama I mean that there are three rbx failures on your PR:
https://travis-ci.org/globalize/globalize/jobs/163254274
https://travis-ci.org/globalize/globalize/jobs/163254275
https://travis-ci.org/globalize/globalize/jobs/163254276

Those are the ones I'm asking about 😄

PR #555, opened 2016-10-01
Get pluck to work on translations table with Rails 5

### LINKED ISSUE #526: regression - pluck not working with translatable field
There was a regression on the master probably on the last commits to make globalize work with ActiveRecord 5.

ActiveRecord 4.2.6
Globalize: 5.0.1

```ruby
PaymentMethod.includes(:translations).pluck(:name) 
#=> ['Paypal', 'Stripe', 'AliPay', 'FastSpring', 'BrainTree']
```

ActiveRecord 5.0
Globalize: master

```ruby
PaymentMethod.includes(:translations).pluck(:name) 
#=> ActiveRecord::StatementInvalid: Mysql2::Error: Unknown column 'payment_methods.name' in 'field list': SELECT `payment_methods`.`name` FROM `payment_methods` LEFT OUTER JOIN `payment_method_translations` ON `payment_method_translations`.`payment_method_id` = `payment_methods`.`id` WHERE `payment_methods`.`id` IN (250, 9435, 46210, 46662, 48421)
```

The query seems correct, the only problem is getting the correct column.

shioyama:
Aim to fix #526. For now just added a failing spec illustrating the issue.

### DIFF
--- a/lib/globalize/active_record/adapter_dirty.rb
+++ b/lib/globalize/active_record/adapter_dirty.rb
@@ -41,7 +41,8 @@ module Globalize
       end
 
       def _reset_attribute name
-        record.send(:restore_attribute!, name) if record.respond_to? :restore_attribute!, true
+        record.send("#{name}=", record.changed_attributes[name])
+        record.send(:attributes_changed_by_setter).send(:except!, name)
       end
 
       def reset
--- a/lib/globalize/active_record/adapter_dirty.rb
+++ b/lib/globalize/active_record/adapter_dirty.rb
@@ -42,7 +42,7 @@ module Globalize
 
       def _reset_attribute name
         record.send("#{name}=", record.changed_attributes[name])
-        record.send(:attributes_changed_by_setter).send(:except!, name)
+        record.original_changed_attributes.except!(name)
       end
 
       def reset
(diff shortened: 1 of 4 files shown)


lib/globalize/active_record/adapter_dirty.rb
```diff
@@ -41,7 +41,8 @@ def clear_dirty
       end
 
       def _reset_attribute name
-        record.send(:restore_attribute!, name) if record.respond_to? :restore_attribute!, true
+        record.send("#{name}=", record.changed_attributes[name])
+        record.send(:attributes_changed_by_setter).send(:except!, name)
```
shioyama: This fixes the issue but is not what I want to keep as a solution since this is very much a private method we are calling. But at least this is a start.

lib/patches/active_model/dirty.rb
```diff
@@ -0,0 +1,3 @@
+ActiveModel::Dirty.class_eval do
+  alias_method :attributes_changed_by_globalize_setter, :attributes_changed_by_setter
```
shioyama: This is how ActiveModel deals with modifying `changed_attributes` internally, by aliasing to another method, so rather than accessing that private method I just did the same thing here.

See: https://github.com/shioyama/rails/blob/72f97e281059bc983eef5bc8915e53249c623dff/activemodel/lib/active_model/dirty.rb#L260

shioyama:
Just for reference, related to frozen `changed_attributes`: https://github.com/mbleigh/acts-as-taggable-on/issues/580

shioyama:
@globalize/core this look good? If so I'll merge.

PR #556, opened 2016-10-01
Drop locale index when dropping table

### LINKED ISSUE #469: `drop_translations_index` should drop locale index too
`create_translations_index` creates two indexes, but `drop_translations_index` only removes one. Both indexes should be removed there.

shioyama:
Fixes #469.

Btw, the specs don't actually check that the method works, because in all DBs we are testing, dropping the table drops the indexes so removing them doesn't matter. But anyway, we should be removing them correctly, and I check that with this change we do that.


shioyama:
@globalize/core Just going to merge this one, it's pretty minor (just removing an index in a migration method).

Issue #458, opened 2015-08-17
Rails 4.2.3, Psql and globalize 5.01 find_by query broke


krisdigital:
Same as: https://github.com/globalize/globalize/issues/423

Hi,

after upgrading to globalize 5.0 and rails 4.2.3, there is a strange error with find_by queries.

Doing something simple as

```ruby
Category.find_by(name: "Name of the category")
```

results in the following query:
```SQL
Category Load (3.1ms)  SELECT  "categories".* FROM "categories" INNER JOIN "category_translations" ON "category_translations"."category_id" = "categories"."id" WHERE "category_translations"."name" = '--- !ruby/object:ActiveRecord::StatementCache::Substitute {}
' AND "category_translations"."locale" = $1 LIMIT 1  [["locale", "de"]]
```

I searched around already but could not find anything, so maybe someone knows a solution. It worked flawlessly before.

```ruby
create_table "categories", force: :cascade do |t|
    t.string   "name",       limit: 255
    t.integer  "parent_id"
    t.integer  "lft"
    t.integer  "rgt"
    t.integer  "depth"
    t.datetime "created_at"
    t.datetime "updated_at"
  end

create_table "category_translations", force: :cascade do |t|
    t.integer  "category_id",             null: false
    t.string   "locale",      limit: 255, null: false
    t.datetime "created_at"
    t.datetime "updated_at"
    t.string   "name",        limit: 255
  end

```

Thanks in advance!

P.S. Normal find(id) queries are working as usual.

FunkyloverOne:
Hi, I guess you should use this:

Category.includes(:translations).find_by(name: "Name of the category")

find(id) queries are working as usual because you aren't translating the id field.

krisdigital:
@FunkyloverOne Hi, thank you for your tip! I already use and posted this solution https://github.com/globalize/globalize/issues/423#issuecomment-133725817

That way you do not have to change all places in your code where you use find_by queries..

FunkyloverOne:
@krisdigital cool! I wasn't too attentive, thanks :)

shioyama:
Ok, after fiddling for a while, I can reproduce this. The reason it is happening is that you use `name` in *both* the `places` table *and* the `place_translations` table. I don't believe we have a note about this anywhere but generally **you should not do this**. I think we once had a spec to check that Globalize will handle the situation gracefully, but it's gone and generally we assume your model does not have the same column in both tables, because it will definitely screw things up.

[...]

PR #449, opened 2015-06-26
Translated attributes as part of the join where

kiliancs:
This is to fix https://github.com/globalize/globalize/issues/423

Issue #423, opened 2015-02-23
#find_by_attribute_name returns nil and produces weird SQL statement


ghost:
Calling `#find_by_attribute_name` on translated object produces following output:

```
2.2.0 :001 > Place.find_by_name 'London'

  Place Load (4.0ms)  SELECT  `places`.* FROM `places` INNER JOIN `place_translations` ON `place_translations`.`place_id` = `places`.`id` WHERE `place_translations`.`name` = '--- !ruby/object:ActiveRecord::StatementCache::Substitute {}\n' AND `place_translations`.`locale` = 'de' LIMIT 1
 => nil

2.2.0 :002 > Place.where(name: 'London').first

  Place Load (4.0ms)  SELECT  `places`.* FROM `places` INNER JOIN `place_translations` ON `place_translations`.`place_id` = `places`.`id` WHERE `place_translations`.`name` = 'London' AND `place_translations`.`locale` IN ('en', 'de')  ORDER BY `places`.`id` ASC LIMIT 1
  Place::Translation Load (0.4ms)  SELECT `place_translations`.* FROM `place_translations` WHERE `place_translations`.`place_id` IN (1)
 => #<Place id: 1, name: "London", created_at: "2013-04-04 09:48:42", updated_at: "2015-02-23 11:51:41"> 
```
- ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux]
- rails 4.2.0
- globalize 5.0.1

dasnotme:
Seems that adding `joins(:translations)` before calling `find_by_attribute_name` fixed the issue.

```
2.2.0 :001 > Place.joins(:translations).find_by_name 'London'

  Place Load (0.6ms)  SELECT  `places`.* FROM `places` INNER JOIN `place_translations` ON `place_translations`.`place_id` = `places`.`id` WHERE `place_translations`.`name` = 'London' LIMIT 1
  Place::Translation Load (0.3ms)  SELECT `place_translations`.* FROM `place_translations` WHERE `place_translations`.`place_id` = 1
 => #<Place id: 1, name: "London", created_at: "2013-04-04 09:48:42", updated_at: "2015-02-23 11:51:41"> 
```

But I wonder if it is an expected behaviour.

parndt:
Yes, because the column doesn't exist on the other model

shioyama:
Hmm... that SQL looks pretty fishy.

parndt:
@shioyama oh you're right!

[...]

Issue #245, opened 2013-06-24
Behavior for translated attributes when the parent table includes translated columns


gaganawhad:
I have a Model, say Post, that has the two translated fields, title and content. The database table for the Post model  has both the columns title and content, which are also in the post_translations table. This setup is a default result of having an app set-up on a locale and then adding a migration that adds the translations table, with the columns that will be translated. 

So the schema would look something like this: 

``` ruby
  create_table :posts, :force => true do |t|
    t.string     :title
.
.
.
  end

  create_table :post_translations, :force => true do |t|
    t.string     :locale
    t.references :post
    t.string     :title
.
.
.
  end
```

My first question is, is this setup supported by globalize3 ? If its not supported, should I be deleting those columns from my parent model after migrating data to the translations table? And if so, should add a note on the README page ? 

If this is schema supported, then the current master isn't doing it what I expect to in this test: 

``` ruby
  test 'return nil when a translated attribute is set as nil when no fallbacks set up' do
    post = Post.create(:title => 'title',  :locale => :en)
    post.update_attributes(:title => nil, :locale => :de)

    post = Post.first
    assert_equal post.title, 'title'

    Globalize.locale = :de
    assert_equal nil, post.title
  end
```

I have the test and the schema set up in this commit: https://github.com/gaganawhad/globalize3/commit/8ea9c9cbe1863745ce5b233ec5b7fa73922ea626

Is there something I am missing  ?

prusswan:
What was the conclusion? I would prefer to fallback to the parent column, so that the parent table remains usable even when the translations table is broken/missing

parndt:
We now don't refer back to the parent table's columns at all.

PR #557, opened 2016-10-01
Add warning if using same columns on table and translations and clean up test data

shioyama:
Looking through old issues, I noticed there are tons that are caused by the confusion around having translated columns in both the translations table and model table. A warning would help warn people that this is not a configuration Globalize supports.

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -49,7 +49,7 @@ module Globalize
       def check_columns!(attr_names)
         if (overlap = attr_names.map(&:to_s) & column_names).present?
           ActiveSupport::Deprecation.warn(
-            ["These columns are present in both model and translation tables: #{overlap.join(', ')}\n",
+            ["These columns are present in both model and translation tables of #{model_name}: #{overlap.join(', ')}\n",
              "Globalize does not support this configuration, expect problems."].join
           )
         end
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -53,6 +53,8 @@ module Globalize
              "Globalize does not support this configuration, expect problems."].join
           )
         end
+      rescue ::ActiveRecord::StatementInvalid
+        warn "Model missing a table: #{model_name}"
       end
 
       def apply_globalize_options(options)
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -4,6 +4,7 @@ module Globalize
       def translates(*attr_names)
         options = attr_names.extract_options!
         # Bypass setup_translates! if the initial bootstrapping is done already.
+        check_columns!(attr_names)
         setup_translates!(options) unless translates?
 
         # Add any extra translatable attributes.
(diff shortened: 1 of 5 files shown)


shioyama:
@globalize/core @globalize/collaborators 

So, after weeding through a lot of issues and switching back and forth between various commits/versions, I've found that lots of issues are caused by having a column with the name of a translated attribute on the model table. We stopped supporting this configuration a while back (I think at the Rails 3->4 transition, [this issue](https://github.com/globalize/globalize/issues/245) discusses the problem) but we didn't officially *say* that we don't support it, and in fact the code *mostly* continued to work with it.

However, issues like #423 and #458, which were *triggered* by changes in ActiveRecord, are actually caused by this configuration. #449 was added to fix this issue, and so far as I can tell it *does* fix the issue, but since there is no regression test nobody really knows what's going on.

On top of this, it was pointed out in #458 that actually as much as we may want to blame users for this, the [migration code](https://github.com/ballistiq/globalize/blob/3695fe57642eebfebc8020ae91324ca55fe76bed/lib/globalize/active_record/migration.rb#L115-L123) for migrating columns from a model table to a translation table **leave the original column on the model table**, which is exactly what is causing these issues. So, in some sense this is a problem of our making...

Given this situation, here's what I'm doing here:
- First, issue a deprecation warning when this configuration is detected, when setting translated attributes. This is important to alert users that they should expect weird things to happen, and should fix the situatoin by renaming or removing the column on the model table.
- Add a regression test to check that e.g. `find_by` works correctly *even with this bad configuration*. I hesitated to add this, but since we've already implemented a fix in #449, we should at least make sure we don't reverse it.
- Other than that one test, remove all columns from our test data that would issue this deprecation warning. There are lots of them, and they make things all the more confusing for anyone looking at the test code.
- Also, (in another PR to come): Rename the original column when migrating data (see #558), so we stop this situation from happening.

I also looked into how we could possibly handle this situation better. There is a nice feature called `ignored_columns` that was [added in Rails 5](https://github.com/rails/rails/pull/21720) which should very cleanly and reliable allow us to ignore the column on the model table. However, it's not available for Rails 4... so I think we should add code to ignore the columns on the model table, but still issue the warning anyway.

Any thoughts? It feels like cleaning up this whole mess would eliminate like half of the serious/hard issues we've seen posted recently.

parndt:
@shioyama outstanding investigation work!
I am in full agreement with your proposed solution.

shioyama:
@parndt Thanks! I think one more thing is to add a note to the README about this, and maybe link to that from the deprecation warning.

lib/globalize/active_record/act_macro.rb
```diff
@@ -45,6 +46,20 @@ def allow_translation_of_attributes(attr_names)
...
       end
 
+      def check_columns!(attr_names)
+        # If tables do not exist, do not warn about conflicting columns
+        return unless table_exists? && translation_class.table_exists?
+
+        if (overlap = attr_names.map(&:to_s) & column_names).present?
+          ActiveSupport::Deprecation.warn(
+            ["You have defined translated attributes with names that conflict with columns on the model table. ",
+             "Globalize does not support this configuration, remove or rename these columns.\n",
```
shioyama: Maybe this should be: "does not support this configuration *anymore*"...

parndt: 👍🏼

shioyama:
I'll merge this sometime tomorrow, any feedback before then will be incorporated :smile:

parndt:
I'm also positive on a note in the README.

shioyama:
Going to merge this for now, we can add a note in the README in a separate PR.

PR #560, opened 2016-10-03
Fix query methods for unsupported table configuration

shioyama:
There have been many issues related to having columns on the model table with names that conflict with translated attributes (i.e. columns on the translations table), e.g. #423 and #458. #449 fixed one issue with `find_by`, but added duplication to the code and made it harder (for me anyway) to understand what is going on.

I added a regression spec in #557 to check that this fix is working, and in this PR I reverted the changes to query methods in #449 and replaced it with a single change, overriding `columns_hash` on the model table to remove translated attributes:

```ruby
def columns_hash
  super.except(*translated_attribute_names.map(&:to_s))
end
```

This fixes the [failing spec](https://github.com/ballistiq/globalize/blob/d972068192c0f74926f1755b4a39938dc4604d60/test/globalize/translated_attributes_query_test.rb#L76).

A more robust way to fix this would be to use [`ignored_columns`](https://github.com/rails/rails/pull/21720), but this is only available in Rails 5, and this issue appears to only manifest itself in Rails 4, so that's not really a solution.

Patching `columns_hash` seems like it might be risky... but all specs are passing, and I like it better than adding more complexity to our existing query method overrides, so I'm thinking this is okay. We could also limit it to Rails 4 only.

Any thoughts?

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -39,6 +39,11 @@ module Globalize
           # Add attribute to the list.
           self.translated_attribute_names << attr_name
         end
+
+        if ::ActiveRecord::VERSION::STRING > "5.0" && table_exists? && translation_class.table_exists?
+          self.ignored_columns += translated_attribute_names.map(&:to_s)
+          reset_column_information
+        end
       end
 
       def check_columns!(attr_names)
--- a/lib/globalize/active_record/class_methods.rb
+++ b/lib/globalize/active_record/class_methods.rb
@@ -7,10 +7,6 @@ module Globalize
         all.merge translation_class.with_locales(*locales)
       end
 
-      def with_where(opts = {})
-        all.merge translation_class.where(opts)
-      end
-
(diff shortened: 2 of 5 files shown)


shioyama:
So in the end what I did was to solve this columns issue in two ways, one for Rails < 5.0 and one for Rails >= 5.0. For Rails < 5.0 I override `columns_hash` as mentioned above. For Rails >= 5.0 I use the `ignored_columns` feature, which should be pretty robust.

I added a test to check that `columns_hash` on the model table does not include the translated attributes, and it passes in all versions so I think this should be ok. Main thing I'm concerned about is side-effects of overriding `columns_hash`, which is a bit hacky (but not terribly so I think).

@globalize/core Anyone want to have a look at this?

shioyama:
@kiliancs I'm reverting part of your fix from #449 so you might want to check that this fixes the issues you and others had. I added a spec which passes so I think it should be fine, but just in case.

shioyama:
Thanks for checking :smile:

Issue #543, opened 2016-08-24
Release Rails 5 compatible Gem


trcarden:
Can someone please cut a rails 5 compatible gem and place it on ruby gems? The last build is from Feb and only supports Rails 4

awaltman:
+1

minifast-julia:
+1

SergeyKolenko:
+1

jonian:
+1

thaneros:
+1

simi:
Better than +1 will be to provide some feedback. Please can you point your projects to master branch and run specs and do some smoke tests to ensure we can release without breaking your applications?

I think there are still some Rails 5 related problems (latest one I'm aware of -> https://github.com/globalize/globalize/pull/546).

Circuit8:
+1

shioyama:
The inverse issue is fixed in #551. What other issues were there?

shioyama:
I've [tagged](https://github.com/globalize/globalize/issues?q=is%3Aissue+is%3Aopen+label%3Arails5) all the issues related to rails5 compatibility and fixed a few of them (e.g. #551, #555).

shioyama:
Ok I don't see any more rails5-specific issues, but we really need some people to use master in their applications to ensure there are no issues before we release this. Let's give a few more days and then if we see no issues I think we can release 5.10.

shioyama:
Sorry seems there's one issue left to fix: #502. There was a fix in #503 that could be used, although no regression spec.

fschwahn:
So, I've been using globalize with rails5 on commit 7d507472e4076c720d798736137eb28d169eb1a1 and everything I was using was working fine. I did a bundle update to commit b44a06dedcb096c3afa8f6b79464fdea917c1480 and suddenly StrongParameters stopped working (I do `model.attributes = params.require(:model).permit(:column)`).  I suppose commit 4ea5cf73184ad441f665219f456f005ee75debf2 is to blame.

dalpo:
@fschwahn thanks for reporting!
Could you open a new issue with more informations about this?

shioyama:
[...]

PR #563, opened 2016-10-06
Fix assign attributes so it behaves like original

shioyama:
Should fix issue reported in #543.

@dalpo @fschwahn Please have a look, I think this should fix the strong parameters issue. I was looking at the source for the wrong version of Rails when I committed 4ea5cf7.

### DIFF
--- a/lib/globalize/active_record/instance_methods.rb
+++ b/lib/globalize/active_record/instance_methods.rb
@@ -12,14 +12,15 @@ module Globalize
       end
 
       def attributes=(new_attributes, *options)
-        return unless new_attributes.is_a?(Hash)
-        assign_attributes(new_attributes, *options)
+        super unless new_attributes.respond_to?(:stringify_keys) && new_attributes.present?
+        attributes = new_attributes.stringify_keys
+        with_given_locale(attributes) { super(attributes.except("locale"), *options) }
       end
 
       def assign_attributes(new_attributes, *options)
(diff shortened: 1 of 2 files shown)


lib/globalize/active_record/instance_methods.rb
```diff
@@ -12,14 +12,15 @@ def attributes
       end
 
       def attributes=(new_attributes, *options)
-        return unless new_attributes.is_a?(Hash)
-        assign_attributes(new_attributes, *options)
+        super unless new_attributes.respond_to?(:stringify_keys) && new_attributes.present?
+        attributes = new_attributes.stringify_keys
+        with_given_locale(attributes) { super(attributes.except("locale"), *options) }
```
shioyama: Note that we could just call `assign_attributes` here, but then if some other gem decided they wanted to override just `attributes=` we would be bypassing that. I figure this level of code duplication is justified to avoid that.

shioyama:
For reference, [here's](https://github.com/shioyama/rails/blob/73521d586981279a99d3ba038d62e2414125df7a/activerecord/lib/active_record/attribute_assignment.rb#L23) the original `assign_attributes` method from Rails 4.2.5.

shioyama:
@globalize/collaborators Going to merge this, it's an important fix since I think currently strong parameters is broken on translated models with `master`. Please have a look when you have a chance.

PR #569, opened 2016-10-19
Get rid of alias_method_chain

parndt:
Instead, we can use Module#prepend

I've opted for the safer version which is to use `:send, :prepend` because this used to be a protected method in Ruby.


ridiculous:
Should we also update the `globalize.gemspec` then, with `spec.required_ruby_version = '>= 2.0.0'`?

parndt:
@ridiculous yes please, are you able to make that change? I think it makes sense to do this anyway.

ridiculous:
Will do 👍

Issue #519, opened 2016-06-21
uniq clause breaks default scope with SQL function


dholdren:
Looks like this PR https://github.com/globalize/globalize/pull/489 added `.uniq`.  But if my default scope is `order("LOWER(name)")` or similar, then `find_by` will fail with: 
`PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list`

dalpo:
@dholdren Could you try if the issue is already present with the latest master?
I pushed this commit that (in thery) should fixed the issue: https://github.com/globalize/globalize/commit/3610db4cb2447b3ea25d64bf0682dc9704623e47

shioyama:
This needs to be fixed before we can release a new version (#543).

shioyama:
@dalpo Could you maybe just add a spec that checks your fix works? Or maybe there is one and I missed it...

dalpo:
@shioyama I encountered the same issue in a my project using postgres... In the next days I would like to start working on. As first step I'm going to create a new PR with a failing spec so we can eventually discuss how to fix it.

dholdren:
unfortunately, that did not fix my bug, here's a rough approximation of my test case and the generated SQL:

```ruby

**Comment by dholdren** [comment]
I do have I18n fallbacks enabled, which is why there's multiple locales in the query I believe

**Comment by dalpo** [comment]
@dholdren I started working on a new PR: https://github.com/globalize/globalize/pull/571
I hope to fix it soon.

**Comment by shioyama** [comment]
@dholdren We can't support direct SQL on translated attributes, so `order("LOWER(name)")` will never work. You will need to directly reference the translations table.

**Comment by Istanful** [comment]
@shioyama 

This solution solves the problem for me when running rails 4.2.10. 

```ruby
module Globalize
  module ActiveRecord
    module ClassMethods
      if ::activerecord::version::string < "5.0.0"
        def self.columns
          super.reject do |column|
            translated_attribute_names.map(&:to_s).include? column.name
          end
        end
      end
    end
  end
end
```
By rejecting the translated attributes from the columns array activerecord won't try to query the corresponding methods. There is already a [similar solution](https://github.com/globalize/globalize/blob/e37c471775d196cd4318e61954572c300c015467/lib/globalize/active_record/class_methods.rb#L6-L14) present in the codebase that modifies the `columns_hash` instead. However in rails 4.2.10 the `columns_hash` is generated from the `columns` method.

[...]

PR #571, opened 2016-10-22
Postgres sorting by translates attributes with fallbacks

dalpo:
Sorting translated attributes with enabled fallbacks is causing an `ActiveRecord::StatementInvalid: PG::InvalidColumnReference: ERROR` on postgres.
#### SQL Error:

```
SELECT DISTINCT, ORDER BY expressions must appear in select list
```
#### Issue history:
- [Added uniq to with_translations to prevent duplicates when fallbacks are defined #489](https://github.com/globalize/globalize/pull/489)
- [Fix: use DISTINCT query only if necessary](https://github.com/globalize/globalize/commit/3610db4cb2447b3ea25d64bf0682dc9704623e47)
- [uniq clause breaks default scope with SQL function](https://github.com/globalize/globalize/issues/519)

### DIFF
--- a/lib/globalize/active_record/query_methods.rb
+++ b/lib/globalize/active_record/query_methods.rb
@@ -48,6 +48,7 @@ module Globalize
           opts
         end
       end
+
       if ::ActiveRecord::VERSION::STRING < "5.0.0"
         def where_values_hash(*args)
           return super unless respond_to?(:translations_table_name)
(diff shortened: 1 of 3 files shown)


dalpo:
@dholdren @shioyama I have fixed the issue related to the `distinct` statement with enabled fallbacks.
Could you review my PR a this point?

But actually there is another problem...
Viewing the code of @dholdren into https://github.com/globalize/globalize/issues/519 he is trying to sort a translated column using a SQL function `lower`.

```
order(arel_table[:name].lower)
```

This isn't actually managed by globalize. I introduced a new failing spec with this use case.
But I'm not shure how to manage it... Maybe parsing the order column with a regexp?
Any suggestions?

shioyama:
Ah... I think we need to draw a line at how far we try to make translated attributes act like normal attributes. Not sure about this one, will have a closer look asap.

dalpo:
@shioyama Yes, I agree with you. Probably is better that every user manage this use case on its codebase, to avoid monkey patching that could introduce bad side effects... 

Anyway let me know what do you think about this PR :smiley:

shioyama:
I don't think we need to support stuff like `order("LOWER(name)")`. If not, do we still need this PR?

shioyama:
I disagree with this merge. Globalize is getting all kinds of complexity in it to handle corner cases that I don't think it should handle. I think we can release without this.

parndt:
@shioyama well, I'm happy to revert this and also revert #489 and release another beta? What do you want to do?

shioyama:
@parndt It's fine, I think releasing was first priority. Let's come back to this once we get some feedback on the release. Thanks again @parndt :smile:

parndt:
@shioyama yeah, I get so little time to allocate that I just thought "well, I'll just get a beta out!" with similar reasoning to yours.  Luckily, it is only a beta! 😄

dalpo:
Hi guys, excuse me if I disappeared, I was lacking of spare time...
I wasn't really sure about about the implementation of this PR me too.

Anyway @shioyama, the purpose of this pull request was to handle a simple SQL order a query with fallback enables under postgresql which wan't possible due the `DISTINCT` statement.

For example: `Model.order(:name)` and NOT stuff like `Model.order("LOWER(name)")`

lazyatom:
The implementation of this actually introduces a bug, which I've fixed in #603 - I'd really appreciate it if you could check that.

PR #578, opened 2016-11-20
Don't create translation record for default_locale on save (rebased)

jtomaszewski:
https://github.com/globalize/globalize/pull/328 rebased to master and working. Only 1 spec is failing - imho its okay to change the spec, but maybe if u prefer to have some backwards compability config option, we could do it as well.

### DIFF
--- a/lib/globalize/active_record/instance_methods.rb
+++ b/lib/globalize/active_record/instance_methods.rb
@@ -143,7 +143,7 @@ module Globalize
       end
 
       def save(*)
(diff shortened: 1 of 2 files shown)


jtomaszewski:
https://github.com/globalize/globalize/pull/328 rebased to master and working. Only 1 spec is failing - imho its okay to change the spec, but maybe if u prefer to have some backwards compability config option, we could do it as well.

Issue #516, opened 2016-06-13
Bad field type for field :payload (:json), should be :string or :text


morgler:
It seems that globalize only works with :string or :text types. Is there a reason why you don't support :json?

My application is a kind of CMS, allowing the host to define content attributes dynamically (e.g. title, slug, body, …). My engine then saves all these attributes in a flexible :json column. In my mind translating :json would not be different from translating :string or :text – it should be completely transparent to globalize.

Btw: here is the "up" part of my migration:

```
def up
  MyCms::Entry.create_translation_table! title: :string, slug: :string, payload: :json
end
```

keithlayne:
This seems like a really arbitrary restriction.  I also want to use JSON with globalize, but allowing just `:json` would still be restrictive.  I want to use `:jsonb` which makes no sense outside of postgres.  I can imagine other ways you could use this gem with other types too.  I think this restriction should be lifted to provide flexibility.

keithlayne:
I just looked at the code and saw #538.  Instead of adding to a list every time someone wants a new (probably db-specific) type to be supported, it probably makes sense to just remove this check.

PR #579, opened 2016-12-06
Remove type check for translated fields

keithlayne:
#516

### DIFF
--- a/lib/globalize/active_record/exceptions.rb
+++ b/lib/globalize/active_record/exceptions.rb
@@ -8,12 +8,6 @@ module Globalize
           super("Missing translated field #{field.inspect}")
         end
       end
-
-      class BadFieldType < MigrationError
-        def initialize(name, type)
-          super("Bad field type for field #{name.inspect} (#{type.inspect}), should be :string or :text")
-        end
-      end
     end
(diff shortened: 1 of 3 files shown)


shioyama:
@globalize/collaborators What do you think? Seems fine to me, but I don't know the background for why the restriction was put in there in the first place.

mingsheng88:
Haven't contributed before but I believe translation should be meant for fields containing text, which was why the text is there? translation wouldn't make sense for doubles, floats, boolean etc etc. I think its just a defensive mechanism

keithlayne:
If you switch your mindset from 'translations' to 'per-locale values' then the restriction on types makes less sense.  I could imagine use cases for lots of types if you look at it that way (including numerics and booleans).

My main argument is that globalize is a library, and therefore should provide the consumer with as much flexibility as possible.  I would like to see the type check go away completely because it doesn't make much sense to me.  On the other hand, there are lots of types that aren't included in the current type check that would make sense to translate (like JSON).  The maintenance overhead for adding types on a one-off basis would be relatively huge.  It's a slippery slope.

parndt:
Goodbye, type check!

Issue #478, opened 2015-12-02
Join table error - UnknownAttributeError: unknown attribute 'locale'


cpilegard:
Using ref: '5db91fc'

```
class CommodityVariety < ActiveRecord::Base
  translates :code, :description, fallbacks_for_empty_translations: true
  ...
  has_many :commodity_varieties_regions
  has_many :countries, through: :commodity_varieties_regions
  ...
  accepts_nested_attributes_for :commodity_varieties_regions
end

class Country < ActiveRecord::Base
  translates :short_name, fallbacks_for_empty_translations: true
  ...
  has_many :commodity_varieties_regions
  has_many :commodity_varieties, through: :commodity_varieties_regions
  ...
  default_scope { order(:short_name) }
end

class CommodityVarietiesRegion < ActiveRecord::Base
  belongs_to :commodity_variety
  belongs_to :country
end
```

If I create a `CommodityVariety` with a `country_ids` attribute, I get the following error:

```
ActiveRecord::UnknownAttributeError: unknown attribute 'locale' for CommodityVarietiesRegion
```

I'm not sure why it's looking for a `locale` attribute on the join table. Any ideas?

cpilegard:
It looks like it's related to this issue: https://github.com/globalize/globalize/issues/145. When I remove the default scope on Country, everything works as normal.

webzorg:
This needs a fix. I spent 15 minutes panicking on what went wrong... I was lucky to find this thread online.

ismaelga:
I got the exact same problem but with just a has_many association. with ref 607e6ef
```ruby

class Project < ApplicationRecord
  translates :name, :pitch
  ...
  has_many :income_shares, dependent: :destroy
  ...
end

class IncomeShare < ApplicationRecord
  translates :label
  ...  
  default_scope { order(value: :desc) }
end
```

Changing from `order(value: :desc)` to `order('value desc')` made the error go away.

**NOTE**: While writing this I tried master it works fine with `order(value: :desc)`

luukveenis:
@shioyama I saw that you tried to reproduce this issue in a failing spec. I'm still seeing this issue in my application and have created a minimal Rails app to reproduce it. I included steps to reproduce the issue in the README:
https://github.com/luukveenis/globalize_test

On my `Bar` model I have a `default_scope` that orders by the translated attribute. The issues seems to be that the default scope includes translations in the query and this AR override includes the locale param into the `where_values_hash`:
[...]

PR #586, opened 2017-01-25
Add failing spec for has-many-through assignment issue

shioyama:
For issue described here: https://github.com/globalize/globalize/issues/543#issuecomment-272989755

And also in #478.


shioyama:
Never mind, these specs are passing.

Issue #210, opened 2013-03-07
Translations update should touch: true master model


batamire:
Translation is defined with belongs_to parent model. It should also default to { touch: true } to make defining cache keys easier.

PR #587, opened 2017-02-06
Add ":touch => true" on belongs_to globalize_model

saiqulhaq:
Parents model.updated_at attr should be updated when globalize attribute has been updated


saiqulhaq:
This patch is applied for version 3-1-stable
previously the bug was reported on https://github.com/globalize/globalize/issues/210, then patched by https://github.com/globalize/globalize/commit/4088241a4facf6125c342430684b0fd97cef626d

parndt:
Thanks 😄

PR #588, opened 2017-02-07
Clone relation that raises immutability exception (fixes #531)

### LINKED ISSUE #531: Deprecation warning with Rails 4.2.7
I'm getting `DEPRECATION WARNING: Modifying already cached Relation. The cache
will be reset. Use a cloned Relation to prevent this warning.` (introduced [here](https://github.com/rails/rails/commit/1b7aa62b184c4410c99208f71b59bbac5c5f03be#commitcomment-9147803)) when my code calls the `.not` method
[here](https://github.com/globalize/globalize/blob/master/lib/globalize/active_record/query_methods.rb#L6-L12) in 5.0.1. I tried cloning the scope and the warning dissapeared:

``` diff
7c7
<           if parsed = @scope.parse_translated_conditions(opts)

---
>           if parsed = @scope.clone.parse_translated_conditions(opts)
```

To be clear, my code never entered the condition. It always went to `super`, but the check is causing the warning. I'm not PRing because I don't fully understand the code and I don't have time right now to dig deeper, but all the tests are green with this change.

mauriciopasquier:
This PR fixes an issue where going through a `has_many :through` relation with a `where.not`
condition triggers a Deprecation warning in rails 4 and raises an exception in rails 5. More details
in the issue #531.

Let me know if the (light) aproach to testing it was ok.

### DIFF
--- a/lib/globalize/active_record/query_methods.rb
+++ b/lib/globalize/active_record/query_methods.rb
@@ -3,7 +3,7 @@ module Globalize
     module QueryMethods
       class WhereChain < ::ActiveRecord::QueryMethods::WhereChain
         def not(opts, *rest)
-          if parsed = @scope.parse_translated_conditions(opts)
+          if parsed = @scope.clone.parse_translated_conditions(opts)
             @scope.join_translations.where.not(parsed, *rest)
           else
             super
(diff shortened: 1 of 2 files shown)


parndt:
Merged: c86d675d46c63d6eb34305ad2e5ffea496b0c61d with 525352dff51ecd1f6384c78e59b2e3850723d6e3

parndt:
Thank you @mauriciopasquier !

mauriciopasquier:
No problem!

PR #589, opened 2017-02-13
fix boolean field lookups for #187

ncri:


ridiculous:
How about a test to prove it fixes something?

ncri:
That's ridiculous. ;-) Kidding. Yes. good point. Have not much time currently unfortunately, also got an error yesterday when trying to run the tests locally.

And indeed, this might have been fixed already in the meantime, as I cannot reproduce the issue anymore I described in https://github.com/globalize/globalize/issues/187. Hm. I will do some more tests later.

ridiculous:
Haha well the reason I ask is because this doesn't appear to change anything; it's exactly the same as before, just a written a lil differently.

ncri:
Yes, very odd. It wasn't the same when I added the patch. It actually did fix something. :) Well I guess the file has been changed in the meantime. I will check the commits.

ncri:
Okay, here we go: https://github.com/globalize/globalize/commit/5d0b8ff976e5223d982cd2c98a13ac0ec3a1ac71 

Has already been fixed quite a while ago. ;-)

PR #598, opened 2017-03-18
Fix for postgres: remove a column only if exists

ghn:
This is a fix for Postgres (haven't tried others) when trying to remove a column that does not exist. 

This is caused by the config `remove_source_columns: true` leading to this error

```
PG::UndefinedColumn: ERROR:  column "title" of relation "books" does not exist
: ALTER TABLE "books" DROP "title"
```


parndt:
Makes sense, thank you! 😄

PR #602, opened 2017-03-22
Check if there's a connection before table_exists? [fix #601]

### LINKED ISSUE #601: rails asset:precompile attempts to connect to DB because of globalize
Hi!

I've upgraded to rails 5.0.2 (from 4.2.7) and using globalized 5.1.0.beta1

When I run `rake assets:precompile` in production mode it throws the error `PG::ConnectionBad: could not connect to server: Connection refused`

And the culprit for what I could find by now is `check_columns!` at https://github.com/globalize/globalize/blob/master/lib/globalize/active_record/act_macro.rb#L49

Is it supposed to run on precompile? What I find odd is that the older version of globalized didn't had this behaviour.

I'm trying now to create a minimal app to reproduce the issue.

ismaelga:
Related to #601

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -40,7 +40,7 @@ module Globalize
           self.translated_attribute_names << attr_name
         end
 
-        if ::ActiveRecord::VERSION::STRING > "5.0" && table_exists? && translation_class.table_exists?
+        if ::ActiveRecord::VERSION::STRING > "5.0" && connected? && table_exists? && translation_class.table_exists?
           self.ignored_columns += translated_attribute_names.map(&:to_s)
           reset_column_information
         end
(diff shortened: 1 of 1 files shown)


ismaelga:
@shioyama I've rebased to master.
I don't think the failing tests are related to my changes but I could be wrong.

parndt:
Thanks @ismaelga !

fsword:
This PR isn't merged yet? I cannot find the changes from master branch

parndt:
See e60b0cffce0083b6f18c71966725b9f08998eb8c

onemanstartup:
@parndt https://github.com/globalize/globalize/pull/619 this pr is broke and table_exists? still making connections.

Bramjetten:
@onemanstartup From #619:
> connected? doesn't return true when running tests using a database because there's nothing to trigger the actual connection. Instead I suggest to simply rescue ActiveRecord::NoDatabaseError to handle the case in #601.

onemanstartup:
@Bramjetten I don't have problems with tests when running with connected is there a problem with globalize tests? I use docker and globalize@master connects to db when I'm running assets:precompile. This is what this pr is all about, and rescue doesn't rescue PG::ConnectionBad error

Tensho:
Confirm, I still have this annoying issue with `PG::ConnectionBad: timeout expired`, raised on `table_exists?` call to the connection object.

@Bramjetten Please could you clarify what's wrong with `connected?`. 🙇

Bramjetten:
@Tensho Looks like something else is going on. We switched to [mobility](https://github.com/shioyama/mobility) a short while after this PR was merged. It's really nice.

PR #603, opened 2017-03-23
Ensure returned records have the correct ID when ordering

lazyatom:
In f1f1628f, the `Arel.star` used, would mean that *all* columns from the source table and joined tables get returned. This meant that there were multiple columns with the name "id". It appears that ActiveRecord will use the latest "id" column to populate the attribute of the returned record.

This is a problem when the IDs of the translation records don't match the ID of the source model, which is almost always the case outside of these simple tests. To demonstrate this, we only need to create a new translation for a model, and then build a query that finds it with that ID-differing locale.

I believe the fix is simple; rather than producing queries of the form `SELECT DISTINCT *, product_translations.name FROM ...`, we just need to only request the source table columns: `SELECT DISTINCT products.*, product_translations.name FROM ...`.

### DIFF
--- a/lib/globalize/active_record/query_methods.rb
+++ b/lib/globalize/active_record/query_methods.rb
@@ -86,7 +86,7 @@ module Globalize
         # Inject `full_column` to the select values to avoid
         # PG::InvalidColumnReference errors with distinct queries on Postgres
         if select_values.empty?
-          self.select_values = [Arel.star, full_column]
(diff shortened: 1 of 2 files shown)


parndt:
Thanks!

dalpo:
Thanks @lazyatom for the great work!

@parndt:
In my opinion this is an important fix, and probably we should consider to release a version soon :smiley:

PR #606, opened 2017-04-03
make touch: true an option

krisdigital:
Makes the touch: true an option. In the past `touch: true` was standard, which was removed because of this rails issue https://github.com/rails/rails/pull/9448, which should be fixed. I also updated the readme.

### DIFF
--- a/lib/globalize/active_record/class_methods.rb
+++ b/lib/globalize/active_record/class_methods.rb
@@ -56,7 +56,11 @@ module Globalize
             klass = self.const_set(:Translation, Class.new(Globalize::ActiveRecord::Translation))
           end
 
-          klass.belongs_to :globalized_model, :class_name => self.name, :foreign_key => translation_options[:foreign_key], inverse_of: :translations, touch: translation_options[:touch] || false
+          klass.belongs_to :globalized_model, 
+            :class_name => self.name, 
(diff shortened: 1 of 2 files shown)


lib/globalize/active_record/class_methods.rb
```diff
@@ -56,7 +56,7 @@ def translation_class
             klass = self.const_set(:Translation, Class.new(Globalize::ActiveRecord::Translation))
           end
 
-          klass.belongs_to :globalized_model, :class_name => self.name, :foreign_key => translation_options[:foreign_key], inverse_of: :translations
+          klass.belongs_to :globalized_model, :class_name => self.name, :foreign_key => translation_options[:foreign_key], inverse_of: :translations, touch: translation_options[:touch] || false
```
parndt: here we could use `fetch`:

```ruby
klass.belongs_to :globalized_model, 
  class_name: self.name, 
  foreign_key: translation_options[:foreign_key], 
  inverse_of: :translations, 
  touch: translation_options.fetch(:touch, false)
```

I find this makes it easier to read and understand that the `|| false` applies to the `touch` option.

krisdigital: Done!

parndt:
Neat, thank you!

krisdigital:
Thank you for merging!

PR #612, opened 2017-04-28
Bump rails dependency versions to '< 5.2'

keithlayne:
This gem has been working fine with the Rails 5.1 RC releases.  Now that 5.1 final is released I assume that globalize will be compatible until at least 5.2.


parndt:
Absolutely - thanks

PR #633, opened 2017-09-04
Fix ActiveRecord serialize default coder values for Rails 5+

AnrichVS:
Fix `Globalize::AttributeMethods::Serialization` to accommodate new ActiveRecord 5+ `Coders::YAMLColumn#initialize` signature. It now takes both the `attr_name` and `class_name_or_coder`, where in ActiveRecord < 5 it only took the `class_name_or_coder`. 

Added a unit test to ensure `Coders::YAMLColumn` serialized column returns correct Array.new default value when DB value is NULL. Updated the 'works with specified marshalling, without data, rails 5.0+' test to reflect changes, both Rails 4 and 5 should return a `Hash.new` in case DB value is NULL and column is serialized as `Hash`.

### DIFF
--- a/lib/patches/active_record/rails4/serialization.rb
+++ b/lib/patches/active_record/rails4/serialization.rb
@@ -1,0 +1,22 @@
+module Globalize
+  module AttributeMethods
+    module Serialization
+      def serialize(attr_name, class_name_or_coder = Object)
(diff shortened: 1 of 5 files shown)


parndt:
Thank you @AnrichVS !

PR #650, opened 2018-01-09
Suggest using the pre-release version for Rails 5

Kjir:
In the README the instructions still suggest to use the master branch from github to use globalize with Rails 5. Now that there are pre-releases, those should be preferred also to make it easier to transition back to a normal release afterwards.


Kjir:
This is now fixed in master, so not needed anymore!

PR #655, opened 2018-01-19
support new ruby versions

ota42y:
- support ruby 2.5.0
- support ruby-head
- upgrade old version's micro version.


dalpo:
thanks

PR #652, opened 2018-01-11
Set MySQL DB encoding to UTF8 in test environment

n-rodriguez:
It should fix errors in Travis : https://travis-ci.org/globalize/globalize/jobs/327791954

```
ActiveRecord::StatementInvalid: ActiveRecord::StatementInvalid: Mysql2::Error: Incorrect string value: '\xE3\x82\xBF\xE3\x82\xA4...' for column 'title' at row 1: INSERT INTO `post_translations` (`locale`, `post_id`, `title`) VALUES ('ja', 1, 'タイトル1')
 ```

PR #651, opened 2018-01-11
Force version of pg gem

n-rodriguez:
https://github.com/rails/rails/pull/31671

PR #656, opened 2018-01-20
jruby test CI fix

ota42y:
jruby activerecord gem support rails 4.2 only during version 1.x only.
So we should freeze version.

When we merge this PR and https://github.com/globalize/globalize/pull/652 and https://github.com/globalize/globalize/pull/651 .
CI will pass.


dalpo:
Thanks

Issue #608, opened 2017-04-04
Bug on count following an order


basex:
The commit https://github.com/globalize/globalize/commit/f1f1628f00fd69c1ad90121496f339299525f890 introduced an error when a `count` is done following an `order`.

Example:
```ruby
products = Product.order(:name)
products.count #=>   DEBUG -   (1.6ms)  SELECT COUNT(*, product_translations.name) FROM `products` INNER JOIN `product_translations` ON `product_translations`.`product_id` = `products`.`id` WHERE `product_translations`.`locale` = 'en'
ActiveRecord::StatementInvalid: Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' product_translations.name) FROM `products` INNER JOIN `product_translations` ON' at line 1: SELECT COUNT(*, product_translations.name) FROM `products` INNER JOIN `product_translations` ON `product_translations`.`product_id` = `products`.`id` WHERE `product_translations`.`locale` = 'en'
```

basex:
I'm having other errors with this commit:
```ruby
Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1: SELECT DISTINCT COUNT(DISTINCT #<struct Arel::Attributes::Attribute relation=#<Arel::Table:0x0000000885aa18 @name="products", @columns=nil, @type_caster=#<ActiveRecord::TypeCaster::Map:0x0000000885aa68 @types=Product(id: integer, price: float)>, @table_alias=nil>, name="*">, product_translations.name) FROM `products` INNER JOIN `product_translations` ON `product_translations`.`product_id` = `products`.`id` WHERE `product_translations`.`locale` IN ('pt', 'en')
```

ckornaros:
I make the same test but with postresql and I have the same issue :(
`products = Product.order(:name)
products.count 
  (9.3ms)  SELECT DISTINCT COUNT(DISTINCT #<struct Arel::Attributes::Attribute relation=#<Arel::Table:0x0055de7fc5d718 @name="products", @engine=Product(id: integer, created_at: datetime, updated_at: datetime, image_id: integer, category: string), @columns=nil, @aliases=[], @table_alias=nil, @primary_key=nil>, name="*">, product_translations.name) FROM "products" INNER JOIN "product_translations" ON "product_translations"."product_id" = "products"."id" WHERE "product_translations"."locale" IN ('fr', 'tig')
[...]

PR #660, opened 2018-02-19
Fix for count following an order

basex:
PG::InvalidColumnReference was corrected in ActiveRecord https://github.com/rails/rails/issues/31783 Commit https://github.com/rails/rails/commit/6df9b69b2363734175d8869393e83e6ce8b4d7b6

Solves https://github.com/globalize/globalize/issues/608


dalpo:
Great! We could finally remove that ugly workaround!

The only think I'm concerned is about the support for older rails versions  (`< 5.1.5`).
@basex Could you add a simple `::ActiveRecord::VERSION::STRING < "5.1.5"` or something like that to mantain the compatibility?

basex:
@dalpo can you merge this PR?

parndt:
I've restarted CI

sergey-alekseev:
This patch actually breaks tests.

PR #664, opened 2018-03-05
Fix: the attribute definition requires an active record cast type arg

dalpo:
Related to: https://github.com/globalize/globalize/commit/e946e639


dalpo:
Duplicated of https://github.com/globalize/globalize/pull/664/files

PR #669, opened 2018-03-13
Fix: Insert translation only when translation has changed  Reverts#341

annaswims:
I would expect that if I don't change any translatable words that no translations will be inserted into the database.  This was the case in version 5.0.1, changed in https://github.com/globalize/globalize/pull/341 for the 5.1.0 realease.  

To give an example of why this is a problem here is our specific use case:
In our application, we have email templates.  We have a set of default templates, and each customer may override the default templates with their own custom version.  The 5.1.0 behavior causes an empty translation record/email to be saved and then we would send out a blank email rather than the default. 

I think if this were merged with https://github.com/globalize/globalize/pull/656  and https://github.com/globalize/globalize/pull/668  then there would only be one failing test as a result of https://github.com/globalize/globalize/commit/e946e63983d42c9ff49b39b13ffcc3ad45626dd3

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -99,11 +99,11 @@ module Globalize
                                 :foreign_key => options[:foreign_key],
                                 :dependent   => :destroy,
                                 :extend      => HasManyExtensions,
-                                :autosave    => true,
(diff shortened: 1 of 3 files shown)


dalpo:
Thanks

PR #675, opened 2018-04-05
Support array as an argument of order

MaicolBen:
So it's possible to use translations in things like `Post.order([:title, :content])`


parndt:
Thanks @MaicolBen

PR #681, opened 2018-04-16
Clean up Ruby environments on Travis CI

deraru:
Hello,

I modified `.travis.yml` and removed some ruby versions which are finished support.
This makes CI time shorter.
Could you review this?

Ref. https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/


parndt:
Thanks, this makes sense

PR #697, opened 2018-08-21
rm an obsolete if condition :scissors:

sergey-alekseev:
ActiveRecord >= 4.2 is currently required:
 https://github.com/globalize/globalize/blob/e37c471775d196cd4318e61954572c300c015467/globalize.gemspec#L19


parndt:
Beautiful use of ✂️

Issue #695, opened 2018-08-02
The name Translation is too generic, we should namespace it


myxoh:
In a project I'm working on, I  am using Globalize por parts of it, but the fact that the module is an un-namespaced module called Translation is causing name conflicts on a new service we are trying to integrate with the project.

It'd be awesome if we, instead, called the module Globalize::Translation (or something like that) to prevent this sort of conflicts.

If I have the time I'll issue a PR, but please do let me know if this is something you'd be open to, as changing all the references might not e that simple

# Activate the gem you are reporting the issue against.
gem 'activerecord', '4.0.0'
gem 'globalize', '5.0.1'
require 'active_record'
require 'globalize'
require 'minitest/autorun'
require 'logger'

# Ensure backward compatibility with Minitest 4
Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :posts, force: true do |t|
  end

  create_table :post_translations, force: true do |t|
    t.references :post
    t.string     :title
    t.text       :content
    t.string     :locale
  end
end

class Post < ActiveRecord::Base
  translates :content, :title
end

class BugTest < Minitest::Test
  def test_association_stuff
    post = Post.create!(title: 'HI')

    assert_equal 'HI', post.title
  end
end

tomash:
This sounds like a good idea. Would you like to make a PR?

myxoh:
Hi @tomash sorry for taking this long to respond. I'll happily write a PR as soon as I have time to re-focus on this. It might take another couple of weeks.

giedriusr:
Sounds like a good idea! 👍

Issue #10, opened 2010-11-19
globalize3 conflicting with i18n database backend


tomash:
the problem seems to come from globalize3 defining Translation model (and #translations association proxy for globalize3-translated classes), while we have this model already for I18n AR backend (according to guide)

here's some gist:
https://gist.github.com/706344

proposed solutions:
1) instead of Translation/translations use Globalize/globalizations (Localization/localizations) by globalize3
2) change I18n guides to advise AGAINST naming i18n translations model "Translation"

PR #701, opened 2018-09-11
Removes class translation defined without namespacing

myxoh:
Addresses: https://github.com/globalize/globalize/issues/695

The class being removed was introduced here: https://github.com/globalize/globalize/pull/12
To solve this issue https://github.com/globalize/globalize/issues/10 

However, removing it doesn't seem to be causing any specs to fail, so either the test being referenced has been removed or this class being defined is no longer required. 

One way or another, having this definition is not a good idea, as it generates many potential conflicts making integration with existing apps more complex.

Before this change calling on a console:
::Translation
=> Results on a defined table
After
::Translation
=> Undefined constant

However, ModelName::Translation, still loads the right table

### DIFF
--- a/test/data/models/translation.rb
+++ b/test/data/models/translation.rb
@@ -1,3 +1,0 @@
-# this little innocent class here makes 0.0.9 test fail
-class Translation < ActiveRecord::Base
-end


myxoh:
I'm curious as whether something change when running the tests on the CI, as they all seemed to be green locally even after removing this class.

PR #705, opened 2018-10-03
Add sqlite3 gem as development dependency

carlesjove:
`sqlite3` is used in tests but not included in the gem list, so `rake test` was raising an error if it wasn't installed in the system.


parndt:
Fair enough - thanks!

Issue #700, opened 2018-09-10
Pluck and various query methods (group, ...) with translated attribute broken on master


kreintjes:
The following code worked on Globalize 5.1.0 (in combination with Rails 5.0 or 5.1) but does not work on the master branch any longer (which is needed for Rails 5.2 support): `Post.with_translations(I18n.locale).pluck(:title)`

I would expect this code to pluck the translated titles automatically (without needing to mention the translations table), similar as is done when ordering or whereing on a translated attribute (which still works).

Instead it returns an unknown column error: `ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: posts.title: SELECT "posts"."title" FROM "posts" INNER JOIN "post_translations" ON "post_translations"."post_id" = "posts"."id" WHERE "post_translations"."locale" = ?`

To reproduce the problem use the following stand-alone test case: https://gist.github.com/kreintjes/dd002db40d23a35b31f7b14c00d21473

The same problem occurs with several [query methods](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html): `group`, `having`, `reorder` and `select`.

Note: in Rails 5.1.6 with Globalize 5.1.0 I get the following deprecation warning which might have something to do with it: `DEPRECATION WARNING: title is not an attribute known to Active Record. This behavior is deprecated and will be removed in the next version of Rails. If you'd like title to be managed by Active Record, add 'attribute :title' to your class. (called from setup at test_globalize.rb:53)`.

However, the problem also occurs in Rails 5.0.4 and 5.1.6 when using Globalize git master. Furthermore, since ordering on a translated attribute (i.e. `order(:title)`) just works, I would expect the same for pluck.

chamnap:
+1

kreintjes:
@basex @chamnap see https://github.com/globalize/globalize/pull/707 for a fix. To use it now (before it's merged): `gem 'globalize', '~> 5.1', github: 'kreintjes/globalize', branch: 'fix/translated-query-methods'`

PR #707, opened 2018-10-16
Add support for translated attributes to other QueryMethods and Calculations

kreintjes:
Adds support to pass the name of translated attribute to various other
QueryMethods and Calculation methods such that Globalize joins the translations
and use the correct database table/column, similar as is done for `order` and
`where`.

Renames QueryMethods to TranslatedAttributesQuery since it also contains methods
not from the ActiveRecord::QueryMethods module (but e.g. from Calculations).
It's also more similar to the test's name now.

Fixes https://github.com/globalize/globalize/issues/700

### DIFF
--- a/lib/globalize/active_record.rb
+++ b/lib/globalize/active_record.rb
@@ -1,14 +1,14 @@
 module Globalize
   module ActiveRecord
-    autoload :ActMacro,        'globalize/active_record/act_macro'
-    autoload :Adapter,         'globalize/active_record/adapter'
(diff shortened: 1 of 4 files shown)


parndt:
@kreintjes thanks - can you please merge in the latest master to this branch?

kreintjes:
@parndt just rebased my branch against master. All specs green now :)

parndt:
Perfect, thank you!

PR #709, opened 2018-10-25
Fix all failed tests

ota42y:
We fixed all failed tests.

**read_attribute bugfix**
Newest Rails use `read_attribute` method and `_read_attribute` method.  
But globalize override `read_attribute` method only so we can't translate 😢 
We override `_read_attribute` method and use translated value.

This method will not be changed, so I am doing a hard copy

**fallback_test bugfix**
Newest I18n exclude default locale  
https://github.com/svenfuchs/i18n/pull/415

So we fix fallback setup data.

### DIFF
--- a/lib/globalize/active_record/instance_methods.rb
+++ b/lib/globalize/active_record/instance_methods.rb
@@ -53,19 +53,18 @@ module Globalize
         end
       end
 
-      def read_attribute(name, options = {}, &block)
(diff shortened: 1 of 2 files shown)


ota42y:
@parndt 
Hi!
By making the test green, I want you to put it in as soon as because it prevents further bugs.

parndt:
Wow, thanks @ota42y and @nabuchi !

PR #711, opened 2018-11-06
FIX: stop registering the attr for Rails 4.2

akfernun:
## Rails Version

4.2.10

## Example 

`Post.includes(:attachments).where(attachments: {id: 123})`

Running the query above in the test suite will raise an exception because `ActiveRecord::Relation#include` relies on `.columns` to help build queries.

## Problem
ActiveRecord uses a model's  `.columns` method to help build queries. For example, `Post.columns` should not include `Post.translated_attribute_names`

### DIFF
--- a/test/globalize/attributes_test.rb
+++ b/test/globalize/attributes_test.rb
@@ -5,11 +5,11 @@ class AttributesTest < MiniTest::Spec
 
   describe 'translated attribute reader' do
     it 'is defined for translated attributes' do
-      if ::ActiveRecord::VERSION::STRING < "5.0"
-        assert_equal Post.new.respond_to?(:title), false
-      else
-        assert_equal Post.new.respond_to?(:title), true
-      end
+      assert Post.new.respond_to?(:title)
+    end
+
+    it 'Post#columns does not include translated attributes' do
+      assert (Post.column_names.map(&:to_sym) & Post.translated_attribute_names.map(&:to_sym)).empty?
     end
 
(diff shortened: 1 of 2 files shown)


akfernun:
@parndt, would be able to review this?

test/globalize/attributes_test.rb
```diff
@@ -5,11 +5,11 @@ class AttributesTest < MiniTest::Spec
...
     it 'is defined for translated attributes' do
-      if ::ActiveRecord::VERSION::STRING < "5.0"
-        assert_equal Post.new.respond_to?(:title), false
-      else
-        assert_equal Post.new.respond_to?(:title), true
-      end
+      assert Post.new.respond_to?(:title)
+    end
+
+    it 'Post#columns does not include translated attributes' do
```
parndt: @akfernun we don't need a version check?

akfernun: @parndt, I don't think so. In Rails 4.2, calling the `attribute` method results in the column being added to the class's `.columns` hash even though it's not an actual column in the database. In Rails 5.2, calling the `attribute` method **does not** result that column being added to the `.columns` hash. Since we added the version check in the code, we can be confident that no matter what version of Rails we're using, calling `.columns` will **never** result in a translated attribute ending up in the `.columns` hash.

McRip:
I would like to see a new release including this fix too.

parndt:
https://rubygems.org/gems/globalize/versions/5.3.0

PR #715, opened 2019-01-29
Add `create_source_columns` option for migrations

IlyasValiullov:
If you use `migrate_date` option on `_create_translation_table_` method, then on roll back, `_drop_translation_table_ `fails because it try to create new columns that exist yet.

So I added create `create_source_columns` property (as reverse for `remove_source_columns`) to `_drop_translation_table_` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Globalize Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


parndt:
@IlyasValiullov thanks for this! Would you mind adding a note in the [README](https://github.com/globalize/globalize/blob/a4b9ad15e9820ce7f585331f5fd8054dc0920228/README.md#migrating-existing-data-to-and-from-the-translated-version) ?

Issue #717, opened 2019-03-01
Rails 6.0 compatibility


cseelus:
Currently installing Globalize with Rails 6 (Beta 2) fails with `globalize (~> 5.2) was resolved to 5.2.0, which depends on activerecord (>= 4.2, < 5.3)`.

Is there anything blocking the use with Rails 6 or is it just not tested yet?

parndt:
Closed by #718 thanks!

PR #718, opened 2019-03-03
Update gemspec to be compatible with Rails 6.0

cseelus:
This PR makes Globalize compatible with Rails 6.0, see https://github.com/globalize/globalize/issues/717

### DIFF
--- a/globalize.gemspec
+++ b/globalize.gemspec
@@ -16,8 +16,8 @@ Gem::Specification.new do |s|
   s.rubyforge_project = '[none]'
   s.required_ruby_version = '>= 2.0.0'
 
-  s.add_dependency 'activerecord', '>= 4.2', '< 5.3'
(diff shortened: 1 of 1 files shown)


morgoth:
Can we have a release with this?

sedubois:
@parndt could you cut a new release to allow using Globalize with Rails 6?

parndt:
Just to note that version 5.3.0 has this https://rubygems.org/gems/globalize/versions/5.3.0

stevenmichaelthomas:
@parndt @svenfuchs would it be possible to create a github release for 5.3.0?

rpbaptist:
Maybe update the changelog as well. Don't want to be rude here, but calling this gem the "Rails I18n de-facto standard library for ActiveRecord model/data translation." I'd expect a little more.

parndt:
@stevenmichaelthomas done - and thanks for the prompt; forgot to tag it.

PR #619, opened 2017-05-17
Support for Rails 5.1

Bramjetten:
There was a change in a private method called `build_relation` in Rails 5.1. I added an updated `uniqueness_validator.rb` in `lib/patches`.

PR #719, opened 2019-03-13
Rescue PG::ConnectionBad on Eager Load

Tensho:
Follow up to #601, #602, #619

### References

- https://github.com/globalize/globalize/pull/602#issuecomment-472343635
- https://github.com/rails/rails/issues/32994


Tensho:
"That's not good for me" – `pg` gem should not become dependency for globalize. I'll try to find the way to sort out the  issue at ActiveRecord level...

PR #721, opened 2019-03-25
Pass the :limit option for the FK column

### LINKED ISSUE #720: Bigint isn't taken as FK
## Problem
There is an issue with translation tables — they're not always taking the correct type of the primary key for their FK types. Here is an example:

```ruby
class CreateProductModels < ActiveRecord::Migration[5.2]
  def change
    create_table :product_models, id: false do |t|
      t.column :id, :bigint, primary_key: true
 # or t.integer :id, limit: 8, primary_key: true
      t.string :brand

      t.datetime :created_at, null: false
    end

    reversible do |dir|
      dir.up do
        ProductModel.create_translation_table! name: :string
      end

      dir.down do
        ProductModel.drop_translation_table!
      end
    end
  end
end
```

This migration produces creation of two tables, where you can see a simple integer for `product_model_id` instead of `bigint`, which is not mentioned in schema.rb:
```ruby
  create_table "product_models", force: :cascade do |t|
    t.string "brand"
    t.string "internal_id"
    t.string "category_tree"
    t.bigint "shop_id"
    t.datetime "created_at", null: false
    t.index ["shop_id"], name: "index_product_models_on_shop_id"
  end

  create_table "product_model_translations", force: :cascade do |t|
    t.integer "product_model_id", null: false
    t.string "locale", null: false
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.string "name"
    t.index ["locale"], name: "index_product_model_translations_on_locale"
    t.index ["product_model_id"], name: "index_product_model_translations_on_product_model_id"
  end
```

## Expected behavior:
I'd expect having a bigint here for the foreign key in the translations table. Otherwise, I should create an additional migration for it.

lancedikson:
## Problem

When the primary key of a translatable table is using bigint type it's not taken as the type of translations' foreign key using `column_descriptor.type`. At first I started considering using `column_descriptor.sql_type`, but there were some issues with it: it contains DB-specific value and seems to be a low-level of getting things done.
The final solution is to take `limit` of a PK and pass it to FK column creation, to make sure they match. It's more API-related way to set things up.

Related issue: #720 

## The PR contains:
- a solution for the problem with having a wrong type for FK of the translation table. In a nutshell, we always check the `sql_type` of a column and check if it contains `bigint` mention (`bigint` for PG and `bigint(8)` for MySQL), and decide what's the type of translation's FK based on that.
- tests, covering the case
- fixes in the testing methods (mysql and psql raw queries using CLI)
- a docker-compose file to make it easier to test MySQL and PG on a local machine
- a fix of sqlite3 version for tests, since it was always failing with the more flexible one

### DIFF
--- a/lib/globalize/active_record/migration.rb
+++ b/lib/globalize/active_record/migration.rb
@@ -78,7 +78,11 @@ module Globalize
 
         def create_translation_table
           connection.create_table(translations_table_name) do |t|
-            t.references table_name.sub(/^#{table_name_prefix}/, '').singularize, :null => false, :index => false, :type => column_type(model.primary_key).to_sym
+            t.references table_name.sub(/^#{table_name_prefix}/, '').singularize,
+                         :null => false,
+                         :index => false,
+                         :type => column_type(model.primary_key).try(:to_sym),
+                         :limit => model.columns.detect { |c| c.name == model.primary_key }.try(:limit)
             t.string :locale, :null => false
             t.timestamps :null => false
           end
--- a/test/globalize/migration_test.rb
+++ b/test/globalize/migration_test.rb
@@ -4,7 +4,7 @@ class MigrationTest < MiniTest::Spec
   include Globalize::ActiveRecord::Exceptions
 
   before(:each) do
-    reset_schema(Migrated, TwoAttributesMigrated)
+    reset_schema(Migrated, TwoAttributesMigrated, MigratedBigint)
     if Globalize::Test::Database.long_table_name_support?
       reset_schema(MigratedWithMegaUltraSuperLongModelNameWithMoreThenSixtyCharacters)
     end
@@ -14,7 +14,7 @@ class MigrationTest < MiniTest::Spec
   end
 
   after(:each) do
-    reset_schema(Migrated, TwoAttributesMigrated)
+    reset_schema(Migrated, TwoAttributesMigrated, MigratedBigint)
     if Globalize::Test::Database.long_table_name_support?
       reset_schema(MigratedWithMegaUltraSuperLongModelNameWithMoreThenSixtyCharacters)
     end
@@ -117,6 +117,11 @@ class MigrationTest < MiniTest::Spec
         # Was it restored? (also tests .untranslated_attributes)
         assert_equal 'Untranslated', untranslated.untranslated_attributes['name']
       end
+    end
+
+    it 'creates a proper types for FK ids' do
+      MigratedBigint.create_translation_table!(:name => :text)
+      assert_migration_table({}, MigratedBigint)
     end
   end
 
@@ -263,7 +268,7 @@ protected
     assert model.translation_class.index_exists_on?(index_field)
 
     assert_equal :string,   column_type(:locale, model)
-    assert_equal :integer,  column_type(index_field, model)
+    assert_equal model.type_for_attribute(model.primary_key), model.translation_class.type_for_attribute(index_field.to_s)
     assert_equal :datetime, column_type(:created_at, model)
     assert_equal :datetime, column_type(:updated_at, model)
 
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ Gemfile*.lock
 .ruby-version
 test/globalize_test.log
 gemfiles/*.lock
+tmp/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,22 @@
+version: '3.7'
+services:
+  postgres:
+    image: postgres:11
+    volumes:
+      - ./tmp/postgres:/var/lib/postgresql/data
+    ports:
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: "postgres"
+      POSTGRES_PASSWORD: ""
+  mysql:
+    image: mysql:8.0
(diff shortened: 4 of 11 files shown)


lancedikson:
Don't really know the reason for the missing sqllite3 dependency, but it's the reason for failing tests.

lancedikson:
Found some issues with using :sql_type, will fix soon. Consider this as WIP for now :)

lancedikson:
Ok, now seems perfect. @parndt, check it out, please, when you have time. I tried to make atomic commits so that it would be convenient to review — each commit contains atomic changes that can be cherry-picked or deleted from the history.
Let me know if there's anything to fix and I'm looking forward to having a release with these fixes 🙏

lib/globalize/active_record/migration.rb
```diff
@@ -72,13 +72,13 @@ def drop_translation_table!(options = {})
         # It's a problem because in early migrations would add all the translated attributes
         def complete_translated_fields
           translated_attribute_names.each do |name|
-            @fields[name] ||= column_type(name)
+            @fields[name] ||= (column_type(name) || :string)
```
parndt: can `column_type` be the one to default to `:string`, if this is still needed, so that the caller doesn't need to know?

lib/globalize/active_record/migration.rb
```diff
@@ -166,7 +166,16 @@ def validate_translated_fields
         end
 
         def column_type(name)
-          columns.detect { |c| c.name == name.to_s }.try(:type) || :string
+          column_descriptor = columns.detect { |c| c.name == name.to_s }
+          column_descriptor_type = column_descriptor.try(:type)
+          column_descriptor_sql_type = column_descriptor.try(:sql_type)
+
+          # return :bigint for the type instead of :integer
```
parndt: could you write _why_ this is happening please?

test/globalize/migration_test.rb
```diff
@@ -263,7 +269,8 @@ def assert_migration_table(fields, model = Migrated)
     assert model.translation_class.index_exists_on?(index_field)
 
     assert_equal :string,   column_type(:locale, model)
-    assert_equal :integer,  column_type(index_field, model)
+    # it's :bigint by default starting from Rails 5
+    assert_equal (Globalize.rails_5? && !Globalize::Test::Database.sqlite?) ? :bigint : :integer, column_type(index_field, model)
```
parndt: this will start failing from `rails_6?` onward :wink:

parndt:
Thanks @lancedikson

lib/globalize/active_record/migration.rb
lancedikson: As you can see, I changed the testing `column_type` to use the same `column_type` just to avoid code duplication, and for tests we need to be able return `nil` in some cases:
https://github.com/globalize/globalize/blob/b007dedab910ce4a1ceb6d071d2ea11a1bb717ae/test/globalize/migration_test.rb#L258-L260

 So, I thought, that it would be better if `column_type` returned either a type or `nil` if it's not existing for some reason.

lib/globalize/active_record/migration.rb
lancedikson: You mean, why typing misconfiguration happening? In a nutshell, before this PR globalize passed a type of `td.reference` during the creation of the translation table, which it took from `column_definition.type` ("Rails"-type), instead of a real data type. Here is an example of those in AR:
https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/activerecord/test/cases/adapters/postgresql/serial_test.rb#L65-L70

I didn't find any proper documentation or comments why exactly AR do split those types and wouldn't return `:bigint` on `cd.type` call instead of `:integer`.
Should I put anything to the comment in the code? Wouldn't like to blow it up with this whole story. We could simply put a link to this thread. What do you think?

test/globalize/migration_test.rb
lancedikson: Oh, right. Didn't think of that. Will fix :)

lancedikson:
Small update, I might have found a better solution. Need to test it carefully. Will text back when got results.

lib/globalize/active_record/migration.rb
parndt: thanks for that!

lib/globalize/active_record/migration.rb
parndt: yes, a link to the pull request at least would be really helpful when trying to figure this code out in the years to come, if we're still doing this :smile: 	thank you!

parndt:
Thanks a lot @lancedikson !

lancedikson:
I changed the way of solving the issue. It seems more reliable and makes more sense now. The point is to pass `:limit` each time we create the FK column for the translations table. It should pass all the tests, I ran them locally on all the DBs and gemfiles.

lancedikson:
@parndt, and now it's perfect... again 😄No drastic changes comparing to what we had, no magic comparisons with `bigint`, so I didn't put any explicit comments to the code as it doesn't seem to be necessary. Check it out again, please, when you have time. Appreciate your work on this project 👍

parndt:
Thanks @lancedikson :+1:

PR #722, opened 2019-04-10
Prevent 'SystemStackError: stack level too deep' error on attribute reset

moiristo:
Prevent 'SystemStackError: stack level too deep' error when resetting a dirty marked attribute that only has values that match the original values. Unchanged, the call loop below occurs for the added test. 

When the original value is reset, the current attribute is still marked dirty, therefore the original value is added over and over again to the old values (dirty) list.

```
/globalize/lib/globalize/active_record/instance_methods.rb:45:in `write_attribute'
/globalize/lib/globalize/active_record/class_methods.rb:96:in `block in define_translated_attr_writer'
/globalize/lib/globalize/active_record/adapter_dirty.rb:49:in `_reset_attribute'
/globalize/lib/globalize/active_record/adapter_dirty.rb:16:in `write'
/globalize/lib/globalize/active_record/instance_methods.rb:45:in `write_attribute'
/globalize/lib/globalize/active_record/class_methods.rb:96:in `block in define_translated_attr_writer'
/globalize/lib/globalize/active_record/adapter_dirty.rb:49:in `_reset_attribute'
/globalize/lib/globalize/active_record/adapter_dirty.rb:16:in `write'
/globalize/lib/globalize/active_record/instance_methods.rb:45:in `write_attribute'
/globalize/lib/globalize/active_record/class_methods.rb:96:in `block in define_translated_attr_writer'
/globalize/lib/globalize/active_record/adapter_dirty.rb:49:in `_reset_attribute'
/globalize/lib/globalize/active_record/adapter_dirty.rb:16:in `write'
/globalize/lib/globalize/active_record/instance_methods.rb:45:in `write_attribute'
/globalize/lib/globalize/active_record/class_methods.rb:96:in `block in define_translated_attr_writer'
/globalize/lib/globalize/active_record/adapter_dirty.rb:49:in `_reset_attribute'
/globalize/lib/globalize/active_record/adapter_dirty.rb:16:in `write'
/globalize/lib/globalize/active_record/instance_methods.rb:45:in `write_attribute'
/globalize/lib/globalize/active_record/class_methods.rb:96:in `block in define_translated_attr_writer'
/globalize/test/globalize/dirty_tracking_test.rb:161:in `block (2 levels) in <class:DirtyTrackingTest>'
```

### DIFF
--- a/lib/globalize/active_record/adapter_dirty.rb
+++ b/lib/globalize/active_record/adapter_dirty.rb
@@ -42,7 +42,9 @@ module Globalize
       end
 
       def _reset_attribute name
-        record.send("#{name}=", record.changed_attributes[name])
+        original_value = record.changed_attributes[name]
+        record.send(:clear_attribute_changes, [name])
+        record.send("#{name}=", original_value)
         record.send(:clear_attribute_changes, [name])
       end
 
--- a/test/globalize/dirty_tracking_test.rb
+++ b/test/globalize/dirty_tracking_test.rb
@@ -154,5 +154,12 @@ class DirtyTrackingTest < MiniTest::Spec
       post.title  = nil
       assert_equal ['content'], post.changed
     end
+
+    it 'only resets attributes once when nothing has changed' do
+      post = Post.create(:title => 'title', :content => 'content')
+      post.send(:set_attribute_was, 'content', 'content')
+      post.content = 'content'
+      assert post.save
+    end
   end
 end


lib/globalize/active_record/adapter_dirty.rb
```diff
@@ -42,7 +42,9 @@ def clear_dirty
       end
 
       def _reset_attribute name
-        record.send("#{name}=", record.changed_attributes[name])
+        original_value = record.changed_attributes[name]
+        record.send(:clear_attribute_changes, [name])
+        record.send("#{name}=", original_value)
         record.send(:clear_attribute_changes, [name])
```
parndt: @moiristo so we definitely need to call this twice?

moiristo: @parndt unfortunately, yes. This first prevents the loop and the second is needed to clear the changes of that assignment. It would not be needed if we could assign using the super assignment method (without dirty tracking). This is actually possible since 2.2 (something like `method("#{name}=").super_method.call(original_value)`. That might be an option when support for < 2.2 is dropped.

parndt: we definitely don't support < 2.4 since 270f0944292894fe762d154d09e8a456b221e1f4

moiristo: Ah, good to know. I see you updated the gemspec already :) I'll check if I can rewrite this then!

moiristo: Actually it is not as easy as I thought. We don't need to call the super method of the assignment, but somewhere it should only use the super method of `def write` instead. I don't think that can be done. The alternative would be to use some flag to skip dirty tracking, but I don't think that would result in a better solution. Therefore, I think we should stick to the original solution.

Issue #723, opened 2019-04-11
rails asset:precompile attempts to connect to DB because of globalize


spikeheap:
- ruby '2.6.1'
- gem 'rails', '~> 5.2.2'
- gem 'globalize', '~> 5.2'

When precompiling assets in a Docker environment, Rails fails with the following error:

```
rails aborted!
PG::ConnectionBad: could not connect to server: Connection refused
  Is the server running on host "localhost" (::1) and accepting
  TCP/IP connections on port 5432?
could not connect to server: Connection refused
  Is the server running on host "localhost" (127.0.0.1) and accepting
  TCP/IP connections on port 5432?
/Volumes/...<omitted>.../app/models/category.rb:18:in `<class:Category>'
/Volumes/...<omitted>.../app/models/category.rb:13:in `<main>'
/Volumes/...<omitted>.../config/initializers/rails_admin.rb:59:in `block in <main>'
/Volumes/...<omitted>.../config/initializers/rails_admin.rb:3:in `<main>'
/Volumes/...<omitted>.../config/environment.rb:7:in `<main>'
/Volumes/...<omitted>.../bin/rails:11:in `<top (required)>'
/Volumes/...<omitted>.../bin/spring:16:in `<top (required)>'
./bin/rails:5:in `load'
./bin/rails:5:in `<main>'
Tasks: TOP => environment
(See full trace by running task with --trace)
```

Because we're doing this as part of a Docker build, we don't have access to any database. Rails is attempting to open a connection because a translation exists on my `Category` model, which is being interrogated by Rails Admin.

This issue originally arose in #601 and was resolved in #602 by guarding database checks with `connected?`. A subsequent PR to support Rails 5.1 (#619) reintroduced this bug in 42b173f66e3a6a63e7908549aa5422881244596e, because `connected?` doesn't return true when running tests using a database because there's nothing to trigger the actual connection. 

#619 altered the behaviour to `rescue ActiveRecord::NoDatabaseError`, however a failed Postgres connection raises `PG::ConnectionBad`, which leads to the error I'm experiencing.

#719 suggested rescuing from `PG::ConnectionBad`, but understandably this was rejected as it makes `pg` a dependency of `globalize`, and doesn't cater for `Mysql2::Error::ConnectionError `.

[...]

PR #724, opened 2019-04-11
Revert "Rescue ActiveRecord::NoDatabaseError..." to enable asset compilation in Docker

spikeheap:
This reverts commit 42b173f66e3a6a63e7908549aa5422881244596e to resolve (or partially work towards) https://github.com/globalize/globalize/issues/723.

I'm not sure if this is acceptable, as there seemed to be good motivations for the switch, but https://github.com/globalize/globalize/pull/602#issuecomment-472346040 (by the author of the patch this reverts) suggested there might have been something else going on. 

I figured I'd create this PR, just in case 😀.

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -40,19 +40,15 @@ module Globalize
           self.translated_attribute_names << attr_name
         end
 
-        begin
-          if ::ActiveRecord::VERSION::STRING > "5.0" && table_exists? && translation_class.table_exists?
-            self.ignored_columns += translated_attribute_names.map(&:to_s)
-            reset_column_information
-          end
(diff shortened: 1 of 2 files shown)


parndt:
Your last commit made the tests fail unfortunately @spikeheap

spikeheap:
I'm no longer working on a project which uses Globalize, and don't seem to be able to find the time to look into this, so I'll close this PR.

We worked around the problem in our project by using a patched version of Globalize with this change because we were up against a tight deadline. Unfortunately this gem in its current state seems incompatible with a Docker-based workflow because of https://github.com/globalize/globalize/issues/723, but there are other ways to approach solving that. My next role is Java-based, but hopefully in the future I'll return to Rubyland and see where this is up to.

parndt:
Thanks @spikeheap

PR #737, opened 2019-09-02
Fix asset precompiling on Docker et al

parndt:
We shouldn't try to connect to the database during asset precompilation, 
for example.

Thanks to #723 and #724

### DIFF
--- a/lib/globalize/active_record/act_macro.rb
+++ b/lib/globalize/active_record/act_macro.rb
@@ -41,7 +41,7 @@ module Globalize
         end
 
         begin
-          if ::ActiveRecord::VERSION::STRING > "5.0" && table_exists? && translation_class.table_exists?
+          if ::ActiveRecord::VERSION::STRING > "5.0" && connected? && table_exists? && translation_class.table_exists?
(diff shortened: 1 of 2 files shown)


albertoalmagro:
We tried this fix and it fixed the issue with assets precompilation.

parndt:
> We tried this fix and it fixed the issue with assets precompilation.

awesome! We just need to get CI passing

flobby:
Would be great if this can be merged back :) globalize is a dependency for a lot of gems that cannot be used thus-for in a docker / asset precompilation scenario

parndt:
@flobby I agree! All that's required is to get CI passing with this change.

thanhnha1103:
> We shouldn't try to connect to the database during asset precompilation.
+1

parndt:
I haven't been able to figure out why the presence of `&& connected?` makes the specs fail here https://github.com/globalize/globalize/pull/737/files#diff-f9375fcce9ef785ba7a6f7fa0f19553fafaf7f89ab69e033f454f7c02284afa0R44

PR #739, opened 2019-11-28
Removing the database connection in initialization + tests

panbanda:
Initial review of a fix for #723 that removes database checks for column names and ignored columns.  This suggestion is based off the fact that it should not be connecting to the database for tests, and before it is in a production setting.  

Developers should _technically_ be reading the docs to see how it uses the columns and translation tables before implementing.  :D .  Shoutout for love of docs <3


panbanda:
@parndt so here is the start of the solution I think.  I don't know how you guys feel about the actual test of migrations and rollbacks.  For example, to test if the data is migrated with `migrated: true` it uses the attr_accessor that is ignored.  However, we could use a DB select that bypasses the model itself to check the value.  Lots of the errors are related to these migration checks.

parndt:
@panbanda bypassing sounds good. It would be nice to solve this issue!

parndt:
@panbanda if you rebase against master or merge in master it'll run CI using GitHub Actions instead

panbanda:
Hey @parndt sorry for the delay.  I have a very slammed Q1 so I probably won't be able to get to this until after that.  If there is someone else willing to jump in happy to turn it over to them.

PR #744, opened 2019-12-13
 Document global `fallbacks_for_empty_translations` setting 

parndt:
This is based on the commits in PR #467, extended by @phansch :smile: 

Closes #467
Closes phansch/globalize#1

### DIFF
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -54,7 +54,7 @@ module Globalize
     end
     
     def fallbacks_for_empty_translations=(value = false)
-      return if !!value != value # Only allow booleans
+      return if [true, false].include? value
       storage[:globalize_fallback_empty_translations] = value
     end
 
--- a/lib/globalize.rb
+++ b/lib/globalize.rb
@@ -52,6 +52,15 @@ module Globalize
     def default_fallbacks(for_locale = self.locale)
       i18n_fallbacks? ? I18n.fallbacks[for_locale] : [for_locale.to_sym]
     end
+    
+    def fallbacks_for_empty_translations=(value = false)
+      return if !!value != value # Only allow booleans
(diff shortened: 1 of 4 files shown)


phansch:
Oh woops, looks like I opened the PR on my fork instead of here :man_facepalming: 

It looks like some tests around the fallbacks are failing now. I'll take a look later today and see what these failures are about.

parndt:
Thanks @phansch

phansch:
The tests seem to pass only when not using the refactorings from 1b5ae4393. It seems, this commit changed behaviour of globalize, because

1) `fallbacks_for` starts returning false for `nil` inputs
2) `[true, false].include? value` doesn't coerce as `!!value != value`, which changed behaviour in some places, too

I pushed a revert of the commit to see if CI passes.

lib/globalize.rb
```diff
@@ -54,7 +54,7 @@ def default_fallbacks(for_locale = self.locale)
     end
     
     def fallbacks_for_empty_translations=(value = false)
-      return if [true, false].include? value
+      return if !!value != value # Only allow booleans
```
parndt: Curious.. how does this improve on `[true, false].include?(value)`? That version only allows booleans too. :smile:

parndt: Or do you mean this restricts what is passed to the method as `value`?

parndt: if that's our intention, perhaps we should just guard:
```ruby
raise ArgumentError, 'fallbacks_for_empty_translations= can only be true or false' if [true, false].exclude?(value)
```

phansch:
Slipped off my radar, I will try and get to this next week :smile:

parndt:
Thanks @phansch

PR #769, opened 2020-12-25
Fix the reference translation

stokarenko:
It fixes [a little regression](https://github.com/globalize/globalize/pull/709/files#diff-d14f092167b54dd486cb43639c5be55de3a92bb741db0066c3415acad4429db3R242) coming in `v5.2`:

1) now we yield the translated value into the block passed to `_read_attribute`, but..
2) in terms of ActiveRecord [this block means the `missing_attribute` callback](https://github.com/rails/rails/blob/v6.0.3.4/activerecord/lib/active_record/associations/belongs_to_association.rb#L119), so it happen that..
3) `globalize` successfully translates, say, `media_id` to `5` and yields the value into `missing_attribute` callback, effectively..
4) raising pretty weird `ActiveModel::MissingAttributeError: missing attribute: 5` exception.

This only happen when we define the foreign reference as translated attribute (don't ask me why.. ;) )

PR #771, opened 2021-01-02
Fixes for Rails 6.0 and 6.1

pixeltrix:
This PR makes the test suite pass locally for me using the SQLite database adapter on Rails 4.2 through to 6.1

I've tried to do the minimum possible to make it work but I have concerns about the changes for 6.0+ since Active Model has tweaked it's internals quite a bit. The Active Model API surface area is very large and I'm pretty sure I'll have missed something related to tracking changes.


parndt:
@pixeltrix thanks for doing this; In parallel, because I noticed CI wasn't working properly in #769, I've fixed CI in #772 by using the Ruby organisation's setup-ruby action, but is that going to make this more difficult for you?

pixeltrix:
@parndt it's fine, I'll just need to tweak this PR - merge #772 and I'll make the change

parndt:
Done - thanks!

pixeltrix:
Updated the workflow changes 👍

parndt:
Thanks a lot @pixeltrix 🎉 

@tomash did you still want to take a look at this?

tomash:
Great work @pixeltrix !

@parndt thanks, it's good. I thought the build errors came from extensive branching to support both pre-6 and 6-plus versions of activemodel, I was wrong. Also, long time no see! 😎

tomash:
I think that after merging that we should update gemspec (and ensure previous versions don't allow installing with AM 6.0, maybe yank or patchlevel update?) and release a new version, @parndt

parndt:
@tomash yep, let's release it as version 6.0.0, it looks good to me too.

Do you mean that after we release it, we should modify the master branch to target newer versions of Rails? Or do you mean that when we push this version we should restrict it to newer versions of Rails and remove the branching?

parndt (approved):

parndt:
Thanks @pixeltrix, I've merged it in and I've released it as 6.0.0 🎉

tomash:
I mean this option:

> Or do you mean that when we push this version we should restrict it to newer versions of Rails and remove the branching?

So just as you did: have Globalize versions follow more-or-less Rails version numbers (at least major number), with simple and unbranched code 😎

PR #773, opened 2021-01-11
[5.3] Fix reference translation

stokarenko:
This is a port of https://github.com/globalize/globalize/pull/769 to `5-3-stable` branch:

> It fixes [a little regression](https://github.com/globalize/globalize/pull/709/files#diff-d14f092167b54dd486cb43639c5be55de3a92bb741db0066c3415acad4429db3R242) coming in `v5.2`:
>
> 1) now we yield the translated value into the block passed to `_read_attribute`, but..
> 2) in terms of ActiveRecord [this block means the `missing_attribute` callback](https://github.com/rails/rails/blob/v6.0.3.4/activerecord/lib/active_record/associations/belongs_to_association.rb#L119), so it happen that..
> 3) `globalize` successfully translates, say, `media_id` to `5` and yields the value into `missing_attribute` callback, effectively..
> 4) raising pretty weird `ActiveModel::MissingAttributeError: missing attribute: 5` exception.
>
> This only happen when we define the foreign reference as translated attribute (don't ask me why.. ;) )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ env:
   - DB=postgres
   - DB=mysql
 
+services:
+  - postgresql
+  - mysql
+
 notifications:
   webhooks:
     - https://webhooks.gitter.im/e/40dc875af26f64f39090
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Globalize Changelog
 
+## 5.3.1 (2021-01-11)
+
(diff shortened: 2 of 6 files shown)


stokarenko:
@parndt Here we go, expected to be v5.3.1 )

parndt:
Wow, I forgot that this version would use different CI. Thanks @stokarenko !

parndt (approved):

stokarenko:
@parndt thank you too! )

PR #788, opened 2021-12-17
Bump Rails to 7.1

ekampp:
This bumps Rails dependency to 7.1

I'm uncertain how high you want to go? Perhaps `< 8`? Or more conservatively `7.1`?

### DIFF
--- a/globalize.gemspec
+++ b/globalize.gemspec
@@ -17,8 +17,8 @@ Gem::Specification.new do |s|
   s.require_path = 'lib'
   s.required_ruby_version = '>= 2.4.6'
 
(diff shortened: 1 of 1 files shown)


mynameisdaniel:
@parndt - bump so it's on your radar for review

parndt (approved):

parndt (changes requested):
Thanks!

Needs CI passing

ekampp:
@parndt, it looks like tests are failing because they can't install parts of OS level dependencies. It doesn't look like an artifact from my code change. Thoughts?

PR #790, opened 2022-01-11
Bump Rails to 7.1; add gemfile and update Appraisals

mynameisdaniel:
Based off https://github.com/globalize/globalize/pull/788

cc: @ekampp

I couldn't install pg on my local instance. Also I added the appraisal gemfile. It should be auto generated.. Checking if this will reduce the CI failures. Any help is welcome.


ekampp:
@parndt, it looks like tests are failing because they can't install parts of OS level dependencies. It doesn't look like an artifact from my code change. Thoughts?

parndt:
Thanks for doing this - looks like we've got a real failure, for example https://github.com/globalize/globalize/runs/4834597364?check_suite_focus=true

PR #799, opened 2022-05-19
Fix asset precompile in docker build workflow

djpremier:
Complementing the work started in PR #737 so that the tests run successfully.

Solve Issue https://github.com/globalize/globalize/issues/723

### DIFF
--- a/test/support/database.yml
+++ b/test/support/database.yml
@@ -9,10 +9,11 @@ mysql:
 
 postgresql:
   adapter: postgresql
-  host: localhost
-  port: 5432
+  host: <%= ENV.fetch('POSTGRESQL_HOST', 'localhost') %>
+  port: <%= ENV.fetch('POSTGRESQL_PORT', 5432) %>
   database: globalize_test
   username: postgres
+  password: <%= ENV['POSTGRESQL_PASSWORD'] %>
   encoding: utf8
 
 sqlite3:
--- a/test/support/database.yml
+++ b/test/support/database.yml
@@ -9,11 +9,11 @@ mysql:
 
 postgresql:
   adapter: postgresql
-  host: <%= ENV.fetch('POSTGRESQL_HOST', 'localhost') %>
-  port: <%= ENV.fetch('POSTGRESQL_PORT', 5432) %>
+  host: <%= ENV.fetch("PGHOST", "localhost") %>
+  port: <%= ENV.fetch("PGPORT", 5432) %>
   database: globalize_test
   username: postgres
-  password: <%= ENV['POSTGRESQL_PASSWORD'] %>
+  password: <%= ENV.fetch("PGPASSWORD", "") %>
   encoding: utf8
 
 sqlite3:
--- a/test/support/database.yml
+++ b/test/support/database.yml
@@ -1,10 +1,10 @@
 mysql:
   adapter: mysql2
-  host: 0.0.0.0
-  port: 3306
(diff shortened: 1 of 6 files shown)


djpremier:
In the first "execution" of the `connected?` function it can happen that Rails does not identify an active connection and with that the result is negative (false), using the function together with the `connection` makes it possible that in the current scope there is no connection, the search function exists globally. Here's a reference:

https://api.rubyonrails.org/classes/ActiveRecord/ConnectionHandling.html#method-i-connection

djpremier:
Awaiting test workflow approval, but in local all is good!

image

And also in Azure CI

image

djpremier:
@parndt Can you review this? 🙂

test/support/database.yml
```diff
@@ -9,10 +9,11 @@ mysql:
 
 postgresql:
   adapter: postgresql
-  host: localhost
-  port: 5432
+  host: <%= ENV.fetch('POSTGRESQL_HOST', 'localhost') %>
+  port: <%= ENV.fetch('POSTGRESQL_PORT', 5432) %>
   database: globalize_test
   username: postgres
+  password: <%= ENV['POSTGRESQL_PASSWORD'] %>
```
parndt, line 16:
```suggestion
  host: <%= ENV.fetch("PGHOST", "localhost") %>
  port: <%= ENV.fetch("PGPORT", 5432) %>
  database: globalize_test
  username: postgres
<% if ENV["PGPASSWORD"] -%>
  password: <%= ENV["PGPASSWORD"] %>
<% end %>
```

Maybe we can use these vars which are documented in the postgresql docs as well: https://www.postgresql.org/docs/current/libpq-envars.html

djpremier: Hi @parndt, thanks for answer, updated!

parndt:
Looks decent @djpremier; we now have duplication in the code which we could solve by extracting a private method like this:

```ruby
def database_connection_possible?
  begin
    # Without a connection tentative, the `connected?` function can responds with a false negative
    ::ActiveRecord::Base.connection
  rescue
    # Ignore connection fail because in docker build hasn't a database connection
    nil
  end

  ::ActiveRecord::Base.connected? && 
    table_exists? && 
    translation_class.table_exists?
end
```

which we then use like this:

```ruby
if Globalize.rails_5? && database_connection_possible?
```

and this:
```ruby
return unless Globalize.rails_42? && database_connection_possible?
```

What do you think?

djpremier:
> Looks decent @djpremier; we now have duplication in the code which we could solve by extracting a private method like this:
> 
> ```ruby
> def database_connection_possible?
>   begin
>     # Without a connection tentative, the `connected?` function can responds with a false negative
>     ::ActiveRecord::Base.connection
>   rescue
>     # Ignore connection fail because in docker build hasn't a database connection
>     nil
>   end
> 
>   ::ActiveRecord::Base.connected? && 
>     table_exists? && 
>     translation_class.table_exists?
> end
> ```
> 
> which we then use like this:
> 
> ```ruby
> if Globalize.rails_5? && database_connection_possible?
> ```
> 
> and this:
> 
> ```ruby
> return unless Globalize.rails_42? && database_connection_possible?
> ```
> 
> What do you think?

Hello, uhm I think it's interesting, it might be useful in the future in a possible expansion of features, I'll change it here.

parndt (approved):

parndt:
I've bumped the version, added this change to the change log, and will release it as `6.2.0` thank you 😄

PR #813, opened 2023-11-04
Fix Compatibility Issue with Audited Gem and Globalize Gem

mohamadzoh:
When attempting to install the Audited gem after the Globalize gem, you may encounter the following error:

ruby
Copy code
undefined local variable or method 'globalize_serialized_attributes' for Audited::Audit:Class
This error occurs because the Audited gem is trying to access the globalize_serialized_attributes method, which is undefined in the context of the Audited gem.

Error Message
 (called from block (2 levels) in require at C:/Ruby32-x64/lib/ruby/site_ruby/3.2.0/bundler/runtime.rb:60)
C:/Ruby32-x64/lib/ruby/gems/3.2.0/gems/activerecord-7.1.1/lib/active_record/dynamic_matchers.rb:22:in `method_missing': undefined local variable or method `globalize_serialized_attributes' for Audited::Audit:Class

### DIFF
--- a/lib/patches/active_record/rails6_1/serialization.rb
+++ b/lib/patches/active_record/rails6_1/serialization.rb
@@ -3,7 +3,7 @@ module Globalize
     module Serialization
       def serialize(attr_name, class_name_or_coder = Object, **options)
         super(attr_name, class_name_or_coder, **options)
-        if defined? globalize_serialized_attributes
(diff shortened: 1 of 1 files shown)


pplant:
+1 on this one

parndt:
Sorry, I'm not really understanding what this does; can this be resolved by including the audited gem before the globalize gem?

PR #816, opened 2023-11-22
Fix ActiveRecord load order

### LINKED ISSUE #815: Globalize load ActiveRecord too early (before initializers)
Adding `gem 'globalize'` to 'Gemfile' in a new Rails app, will mean setting active_record configuration in initializers won't work.

Example
```sh
gem install rails
rails new example-globalize-rails-app --minimal
cd example-globalize-rails-app
echo "gem 'globalize'" >> Gemfile
bundle

echo "Rails.application.config.active_record.collection_cache_versioning = false" >> config/initializers/new_framework_defaults_6_0.rb
echo "raise if ActiveRecord::Base.collection_cache_versioning == true" >> config/initializers/new_framework_defaults_6_0.rb

bin/rails runner 1
```

Maybe the code in https://github.com/globalize/globalize/blob/278b9f95feef3838d434226c8a3836cb4e7880ec/lib/globalize.rb#L125-L128 needs to be wrapped in
```
ActiveSupport.on_load(:active_record) do
  # your code
end
```

https://guides.rubyonrails.org/engines.html#configuration-hooks

runephilosof-abtion:
Fixes #815
> Globalize load ActiveRecord too early (before initializers)

Which meant that setting active_record configuration in initializers did not work.


parndt:
This makes sense - thanks @runephilosof-abtion

PR #819, opened 2024-05-10
Adjust serialization patch

intale:
Rails v7.1 introduced changes to the [serialize](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize) method. It now accepts different set of arguments. I reviewed the serialize patch a bit and made it unique across all rails versions.


parndt:
Hi thank you for this - I've merged #824 and #825 which also touch this area, in case that helps with this?

PR #821, opened 2024-08-31
Allow custom `:dependent` option for translations association

brunodccarvalho:
In #48 (a decade ago) the `:dependent` value was changed to `:destroy` for sensible reasons, but without alternative, this is a considerable performance downgrade when in the vast majority of cases the translation model doesn't own anything, has no destroy callbacks, and is not even loaded when the record gets destroyed. So forward the dependent option here.


lib/globalize/active_record/act_macro.rb
```diff
@@ -99,7 +99,7 @@ def setup_translates!(options)
 
         has_many :translations, :class_name  => translation_class.name,
                                 :foreign_key => options[:foreign_key],
-                                :dependent   => :destroy,
+                                :dependent   => options[:dependent] || :destroy,
```
parndt, line 102:
```suggestion
                                :dependent   => options.fetch(:dependent, :destroy),
```

parndt (approved):

parndt:
Thanks @brunodccarvalho

CHANGELOG.md
```diff
@@ -2,11 +2,12 @@
 
 ## Unreleased
 
+* Allow custom :dependent option for translations association [#821](https://github.com/globalize/globalize/pull/821) by [Bruno Carvalho](https://github.com/brunodccarvalho)
```
parndt, line 5:
```suggestion
* Allow custom `:dependent` option for translations association [#821](https://github.com/globalize/globalize/pull/821) by [Bruno Carvalho](https://github.com/brunodccarvalho) 
```

parndt (approved):

PR #822, opened 2024-11-02
1 - обновление доступной версии activerecord

Buruat:


parndt:
This has been resolved by #825 - thank you!

PR #826, opened 2024-11-27
Add for Rails 8.0 and Ruby 3.3 to the test matrix

rocket-turtle:


parndt:
Awesome, thank you @rocket-turtle !

PR #832, opened 2025-04-27
Do not create empty current translation

pyromaniac:
Hello, I noticed that Globalize creates an empty translation for current locale on save (due to a call to `translation`).
I believe this is unnecessary. Also, it fails to save if there are validations on translation_class.

In the previous implementation `translation.locale` is always `Globalize.locale`. 
And `Globalize.locale` will be always set, so `|| I18n.default_locale` part didn't make sense either. 
Thus the `Globalize.with_locale` block call would not do anything except initializing the translation instance.

Also, it was happening only on `save` and not on `save!`. The PR addresses the inconsistency by patching both methods.

### DIFF
--- a/test/globalize/attributes_test.rb
+++ b/test/globalize/attributes_test.rb
@@ -319,12 +319,9 @@ class AttributesTest < Minitest::Spec
       ActiveRecord::JDBCError
     )
 
-    it 'does not save a record with an empty required field' do
-      err = assert_raises ActiveRecord::StatementInvalid do
-        Artwork.create
-      end
-
-      assert_match(/#{DB_EXCEPTIONS.join('|')}/, err.message)
+    it 'does not create an empty translation record' do
+      artwork = Artwork.create
+      assert_equal 0, artwork.translations.length
     end
 
     it 'saves a record with a filled required field' do
--- a/test/globalize/attributes_test.rb
+++ b/test/globalize/attributes_test.rb
@@ -324,21 +324,31 @@ class AttributesTest < Minitest::Spec
       assert_equal 0, artwork.translations.length
     end
 
-    it 'saves a record with a filled required field' do
+    it 'does not save a record with an empty required field' do
+      err = assert_raises ActiveRecord::StatementInvalid do
+        Artwork.create(title: 'foo')
+      end
+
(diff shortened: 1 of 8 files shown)


pyromaniac:
Hello @parndt. Any change this can be merged? Just curious. Or maybe there is a different direction you would like to move?

test/globalize/attributes_test.rb
```diff
@@ -319,12 +319,9 @@ class AttributesTest < Minitest::Spec
       ActiveRecord::JDBCError
     )
 
-    it 'does not save a record with an empty required field' do
-      err = assert_raises ActiveRecord::StatementInvalid do
-        Artwork.create
-      end
```
parndt, line 325: Is there a reason this test was dropped?

parndt:
Thanks for the reminder - I think this change makes sense; ~could you add a note to the unreleased section of the [CHANGELOG.md](https://github.com/globalize/globalize/blob/main/CHANGELOG.md)?~

Actually, I can easily do that.

test/globalize/attributes_test.rb
pyromaniac: I don't remember of course :) Let me find it out.

parndt: Haha, more than fair! Sorry it took a while.

pyromaniac: Ok, found it: the old test asserted that Artwork.create raises ActiveRecord::StatementInvalid because the empty translation row violated the not-null constraint. This PR's change is precisely what makes that no longer happen — save_translations! now skips the current-locale stash when it has no attributes, so no INSERT INTO artwork_translations is issued and the DB never gets a chance to reject it.
But give me a moment to think, maybe it is still possible to have this spec.

pyromaniac: To preserve the original intent (verifying the not-null safety net still works when a translation IS being persisted but is incomplete), I added a second null: false column subtitle to artwork_translations, made Artwork translate both, and restored the test.

pyromaniac:
@parndt hi again, just a friendly reminder :) Or do you have doubts?

parndt:
@pyromaniac no doubts, it's just free time related! Thanks for the reminder, they are helpful.

PR #833, opened 2025-07-16
Fix automatic creation of empty stash entries for unassigned locales

jules-w2:
# Fix automatic creation of empty stash entries for unassigned locales

## Problem Description

When using Globalize with `globalize-accessors`, the gem automatically creates empty stash entries for all available locales, even when no values have been explicitly assigned to those locales. This leads to unexpected behavior where the current locale (`I18n.locale`) appears in the stash and can cause unwanted translation records to be created during save operations.

## Root Cause

The issue occurs in the `contains?` method in `lib/globalize/active_record/attributes.rb`:

```ruby
def contains?(locale, name)
  self[locale].has_key?(name.to_s)  # This line creates empty entries
end
```

When `self[locale]` is called, it triggers the `[]` method which automatically creates an empty hash `{}` for any locale that doesn't exist:

```ruby
def [](locale)
  locale = locale.to_sym
  self[locale] = {} unless has_key?(locale)  # Creates empty entry
  self.fetch(locale)
end
```

### Example: Basic Usage
```ruby
class BlogCategory < ApplicationRecord
  translates :name, :slug
  globalize_accessors
end

# Set locale to French
I18n.locale = :fr

# Create and assign values
b = BlogCategory.new
b.name_en = "Test en"
b.name_it = "Test it"

# Expected stash: {en: {"name" => "Test en"}, it: {"name" => "Test it"}}
# Actual stash: {en: {"name" => "Test en"}, it: {"name" => "Test it"}, fr: {}}
```

## Impact

1. **Performance**: Unnecessary memory usage with empty stash entries
2. **Data Integrity**: Unwanted translation records in database
3. **Confusion**: Difficult to distinguish between explicitly assigned and automatically created translations
4. **Validation Issues**: Can interfere with validation logic that expects only explicitly assigned translations

## Solution

Modify the `contains?` method to check for locale existence before accessing it:

```ruby
def contains?(locale, name)
  has_key?(locale) && self[locale].has_key?(name.to_s)
end
```

This change ensures that:
- Empty stash entries are not created during existence checks
- Only locales with explicitly assigned values appear in the stash
- No unwanted translation records are created during save operations

---

**Files Changed:**
- `lib/globalize/active_record/attributes.rb` - Modified `contains?` method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Globalize Changelog
 
+## unreleased
+
+* Fix automatic creation of empty stash entries for unassigned locales [#833](https://github.com/globalize/globalize/pull/833) by [Jules](https://github.com/jules-w2)
+
 ## 7.1.0 (2025-12-23)
 
 * Add support for Rails 8.1.x [#838](https://github.com/globalize/globalize/pull/838) by [Jules](https://github.com/jules-w2)


jules-w2:
@parndt 
Thanks for the merge.
Do you have an ETA for releasing this change?

parndt:
@jules-w2 probably today 😄

parndt:
@jules-w2 https://github.com/globalize/globalize/releases/tag/v7.1.1

PR #834, opened 2025-08-26
Fixed migration rollback not working when `:migrate_data` option is true

jklimke:
The current implementation of the method that moves translated attributes back to the source table was modifying the translated_attribute_names within the loop over all records. This caused the execution to break on the n + 1st iteration when translated_attribute_names has n entries.
 
as deleting the translated attributes there is not really necessary, i moved it to the end of the overall rollback.


parndt:
CI seems pretty unhappy with this, unfortunately.

PR #843, opened 2026-05-14
Fix Rails 7.2 serialization compatibility

jatin-rajput829:
## Summary

Fixes a load-order issue in Rails 7.2 where `Globalize` can raise:

`NameError: undefined local variable or method 'globalize_serialized_attributes'`

This happens when another gem (e.g. `audited`) is loaded before Globalize and calls `serialize` from its own `ActiveSupport.on_load(:active_record)` hook. In that ordering, Globalize’s patched `serialize` can run before `globalize_serialized_attributes` is initialized.

### DIFF
--- a/lib/patches/active_record/serialization.rb
+++ b/lib/patches/active_record/serialization.rb
@@ -1,15 +1,7 @@
-# `on_load(:active_record)` hooks run in registration order, so a gem
-# required before Globalize (e.g. `audited`) can define a model -- and
-# therefore call `serialize` -- from its own hook before any later Globalize
-# hook runs, reaching the `serialize` patch below while
-# `globalize_serialized_attributes` doesn't exist yet:
-#
-#   NameError: undefined local variable or method `globalize_serialized_attributes'
-#
-# Setting the attribute up here instead, unconditionally as this file loads,
-# guarantees it always exists before the patch does, regardless of gem load
-# order. `active_record` is already required by lib/globalize.rb by this
-# point, so `ActiveRecord::Base` is available without needing `on_load` at all.
+# Gem load order is not guaranteed, so another gem's on_load(:active_record)
+# hook can call `serialize` on a model before this one runs. Defining the
+# attribute unconditionally here, rather than inside a hook, ensures it
+# always exists before that happens.
 ActiveRecord::Base.class_attribute :globalize_serialized_attributes, instance_writer: false
 ActiveRecord::Base.globalize_serialized_attributes = {}
 
--- a/lib/patches/active_record/serialization.rb
+++ b/lib/patches/active_record/serialization.rb
@@ -1,33 +1,17 @@
-module Globalize
-  module AttributeMethods
-    # Storage for the serialization options Globalize replays onto translation
-    # classes in +ActMacro#enable_serializable_attribute+.
-    #
-    # These accessors are deliberately defined here, in a module prepended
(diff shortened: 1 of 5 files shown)


parndt:
Hello! Thanks for the patch; can you please contribute a failing spec that this solves?

rocket-turtle:
I think the call side is not correct here. `coder` should be a kw argument in Rails 7.2

https://github.com/rails/rails/blob/7-2-stable/activerecord/lib/active_record/attribute_methods/serialization.rb#L183

parndt:
I'm going to modify this PR.

parndt:
@jatin-rajput829 @rocket-turtle what do you think?

jatin-rajput829:
Thanks for updating this, @parndt this looks good to me now.
Moving globalize_serialized_attributes into the serialization patch avoids the on_load ordering race, and the new serialization_load_order_test covers the exact failing scenario I hit.
I’m 👍 to merge from my side.

parndt:
Thanks @jatin-rajput829 !

lib/patches/active_record/serialization.rb
```diff
@@ -1,3 +1,34 @@
...
+    # -- and therefore call +serialize+ -- from its own hook, reaching the
+    # patched +serialize+ before Globalize's hook has run. That raised
+    # "NameError: undefined local variable or method
+    # `globalize_serialized_attributes'". Defining them next to the patch makes
+    # it impossible for one to exist without the other.
+    module SerializedAttributes
+      def globalize_serialized_attributes
+        @globalize_serialized_attributes ||=
+          if superclass.respond_to?(:globalize_serialized_attributes)
+            superclass.globalize_serialized_attributes.dup
```
rocket-turtle: The serialization patch also call `.dup` on `globalize_serialized_attributes` I think that it can be removed there or?

https://github.com/globalize/globalize/blob/main/lib/patches/active_record/rails7_2/serialization.rb

That would also remove unnecessary duplications.

rocket-turtle (commented):
I like the patch. With that in place i think we can make an improvment to the [rails7_2/serialization.rb](https://github.com/globalize/globalize/blob/main/lib/patches/active_record/rails7_2/serialization.rb) as well.

lib/patches/active_record/serialization.rb
parndt: @rocket-turtle nice, I've done this just now with 60d1324

rocket-turtle:
@parndt anything left to do?

if you merge this in, i can rebase my Rails 8.1 PR (https://github.com/globalize/globalize/pull/848) on top of it.

parndt:
@rocket-turtle I've just made one change for when this all happens and how: a443b11af91900817efa6f1c742975ac6ae6da75

rocket-turtle:
This looks like a small and good change to fix the described Problem.

Personaly I dont like this big AI comments which explain how we came to that code and not what this code does. If you are new to the code base and read the comment in _lib/patches/active_record/serialization.rb_ it does not help you understand the current code. It will propably just be confusing because you need to understand the history.

The comment in _test/globalize/serialization_load_order_test.rb_ looks good.

parndt:
@rocket-turtle 100% agree with you there. I've updated it.

parndt:
@rocket-turtle thanks for your help!
@jatin-rajput829 thanks for your patch and PR! 

It's merged, so #848 can now be rebased on top of this.

PR #844, opened 2026-05-16
Coerce locale to symbol in Attributes#contains?

pyromaniac:
## What

Convert `locale` to `Symbol` at the top of `Attributes#contains?`, matching the coercion already done in `Attributes#[]` (and therefore in `#read`/`#write` which route through it).

## Why

#833 rewrote `contains?` to skip the `self[locale]` auto-vivification — which is a real fix — but in doing so dropped the implicit `to_sym` that `[]` performs:

```ruby
def [](locale)
  locale = locale.to_sym       # coerce
  self[locale] = {} unless has_key?(locale)
  self.fetch(locale)
end

def contains?(locale, name)
  key?(locale) && self[locale].key?(name.to_s)   # 7.0.0: routed via [], coerced. 7.1.x: not coerced.
end
```

The stash keys are always symbols because every write goes through `self[locale]`. After 7.1.0, calling `contains?("en", "title")` returns `false` even when `:en` is present.

This bites real callers — most visibly [globalize-accessors](https://github.com/globalize/globalize-accessors), which captures whatever was passed to `locales: …` in the macro and forwards it to `stash.contains?` from the generated getter. Something like:

```ruby
translates :title
globalize_accessors locales: I18n.available_locales.map(&:to_s)
```

silently breaks: `post.title_en` reads from the translation table instead of the in-memory stash, so writes made before save aren't visible. This is the pattern in our codebase and is what surfaced the regression for us.

## Repro

```ruby
stash = Globalize::ActiveRecord::Attributes.new
stash.write(:en, 'title', 'foo')

stash.contains?(:en, 'title')   # 7.0.0: true   | 7.1.x: true
stash.contains?('en', 'title')  # 7.0.0: true   | 7.1.x: false  ← regression
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Globalize Changelog
 
+## Unreleased
+
+* Coerce locale to symbol in `Attributes#contains?` [#844](https://github.com/globalize/globalize/pull/844) by [Arkadiy Zabazhanov](https://github.com/pyromaniac)
+
 ## 7.1.1 (2025-12-24)
 
 * Fix automatic creation of empty stash entries for unassigned locales [#833](https://github.com/globalize/globalize/pull/833) by [Jules](https://github.com/jules-w2)
--- a/lib/globalize/active_record/attributes.rb
+++ b/lib/globalize/active_record/attributes.rb
@@ -11,6 +11,7 @@ module Globalize
       end
 
       def contains?(locale, name)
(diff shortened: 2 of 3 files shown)


parndt (approved):
Thanks @pyromaniac

pyromaniac:
Hello @parndt,

Thanks for merging this. Would it be possible to have the release any time soon?

PR #792, opened 2022-02-09
Patch ActiveSupport in Rails 7.0.0+

artisr:
Based off 
https://github.com/globalize/globalize/pull/788
https://github.com/globalize/globalize/pull/790

cc: @ekampp @parndt

This is what failed for us: https://github.com/rails/rails/blob/v7.0.2/activesupport/lib/active_support/inflector/inflections.rb#L68

added ```I18n.respond_to?(:fallbacks)``` before ```I18n.fallbacks[locale].each do |k|```

Issue #847, opened 2026-07-06
instance_or_fallback patch breaks pluralization on Rails 8.1


rocket-turtle:
Rails 8.1 moved the `:en` inflections out of the internal `@__instance__` map into a separate `@__en_instance__` ivar (rails/rails@2ab34cd5). The override in _lib/patches/active_support/inflections.rb_ still looks up locales in @__instance__ only, so with
globalize loaded the `:en` inflections are never found there.

This bit us after upgrading to 8.1: our app defines empty inflections for non-English locales (so English rules don't get applied to German words). `I18n.fallbacks[:en]` includes `:de`, so the patch now returns the empty `:de` instance instead of the `:en`
rules. Result: `"translation".pluralize returns "translation"`, Active Record derives singular table names and the app fails to boot with Could not find table `'translation'`.

Repro (globalize 7.1.3 / rails 8.1.3, fine on ≤ 8.0 and without globalize):

```
require "active_record"
require "globalize"
require "i18n/backend/fallbacks"

ActiveSupport::Inflector.inflections(:de) { }
I18n.fallbacks = I18n::Locale::Fallbacks.new(en: :de)

ActiveSupport::Inflector.pluralize("translation", :en)
# => "translation" (expected "translations")
```

The patch was added in #792 to guard against I18n.fallbacks not being defined (#788). It only needs to keep that guard and can delegate the rest to Rails, which handles its own internals correctly (Rails fixed the same fallback lookup on their side in rails/rails@4f90c986):

```
def instance_or_fallback(locale)
  return instance(locale) unless I18n.respond_to?(:fallbacks)

  super
end
```

The globalize suite passes with this on 7.2–8.1, and we've also been running our application's test suite against the patched gem without issues.

Note: vanilla Rails 8.1.0/8.1.1 have the same broken fallback lookup themselves (fixed in 8.1.2 by rails/rails#56344), so the regression test skips on those two versions.

rocket-turtle:
PR in https://github.com/globalize/globalize/pull/848/commits

rocket-turtle:
fixed on main

PR #848, opened 2026-07-06
Fix inflections with Rails 8.1

rocket-turtle:
Fixes the pluralization regression on Rails 8.1 introduced by #792's `instance_or_fallback` patch.

On Rails 8.1 the `:en` inflections moved out of `@__instance__`, per rails/rails@2ab34cd5, so the fallback lookup silently returned empty inflections: pluralize returned singular words, and Active Record derived singular table names from them. This keeps the `I18n.fallbacks` guard from #788 and otherwise defers to ActiveSupport's original implementation.


rocket-turtle:
@parndt I did some cleanups (removed old rails versions and workarounds) and fixed a bug which I had while upgrading to rails 8.1 (https://github.com/globalize/globalize/issues/847)

I also ran it against our Rails 8.0 testsuit.

I don't know how you stand to AI commits but i checkt everything and it should be good to go.

The inflection bug is a bit confussing because there where also some problems in rails to get there change right (fixed in 8.1.2 by https://github.com/rails/rails/pull/56344).

rocket-turtle:
@parndt I updated the branch and also let it run against our test suit. all green :)

n7st:
We're seeing the same issue during a Rails 8.1 upgrade, and this branch looks to fix it!

parndt:
@rocket-turtle I've cut down the size of the PR to just the inflections fix. I've backed up the original branch here: https://github.com/globalize/globalize/tree/backup/rails-8-1-fix-inflections-848

OK with you? Feel free to amend the commit here if you want to sign it.

parndt:
@n7st thanks - confirming the cherry-picked version of this branch still works for you?

n7st:
The cherry-picked version looks to work well to me.

rocket-turtle:
@parndt looks good to me. I will make a new PR for the cleanup after the merge.

I still think the cleanup is a good idea, reducing the gem complexity.

PR #849, opened 2026-09-02
Require rails 8 0

rocket-turtle:
@parndt here the "promised" cleanup commit.

Maybe you can release a 7.1.4 with the current Unreleased changes and make a 8.0 with this changes which would nicly match the rails version :)

Rails 7.2 - https://endoflife.date/rails
Ruby 3.1 - https://endoflife.date/ruby

Ruby 3.2 is also EOL but I would keep that (don't know why maybe because i think people update rails but might be behind on ruby)

I removed Rails 7.2 as well because if your not yet on Rails 8.0 you could still go with the 7.1 version. There are no features or security thinks on globalize you would miss.


rocket-turtle:
@paneq could you take a look at this PR when you get a chance? Would appreciate your review 🙏

And cut a new version so people can continue with there Rails update. Rails 8.0 EOL is around the corner. 😄

